PR #42, opened 2014-12-23
recursively expands ${VARIABLES} within the value (w/o refactor)

### LINKED ISSUE #39: Variable expansion only works when it matches the entire value
In a bash script, it is common to see things like this:

DEPLOY_TARGET=stage
API=http://api.${DEPLOY_TARGET}.foo.com/

It looks like _processForPotentialEnvVariable does this (w/o the braces), but only if the entire value matches another env entry.  I'm not sure why anyone would want to do that, but filling out the feature to replace variables within the value would be useful.

apm:
My editor ate trailing spaces, but it should be clean enough.  Fixes #39.


motdotla:
Looking into this now

jcblw:
@motdotla  he lives

motdotla:
I have some of the same sentiments as @maxbeatty here, but ultimately I think it is a good PR with additional usefulness.

We're gonna roll with it.

I'll version bump it.

And additionally, I'm going to refactor the tests into individual scenarios for each variable permutation - for added clarity.

motdotla:
After digging deeper into this, I agree with @maxbeatty. This is needlessly complex - especially when there are more straightforward approaches to solving the problem that inspired this pull request.

See my comment here: https://github.com/motdotla/dotenv/issues/39

I'm reverting these changes, and keeping this closed.

PR #48, opened 2015-01-28
set env vars even if environment env file is missing

### LINKED ISSUE #47: .env and .env.environment both required?
Why are both a `.env` and a `.env.ENVIRONMENT` required? Shouldn't just a `.env` file be necessary?

Also, this isn't clear in the README that this is the behavior and threw me for a loop.

To me, this should be a `||` not a `&&`: https://github.com/motdotla/dotenv/blob/master/lib/main.js#L101

maxbeatty:
fixes #47

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ Add it to your package.json file.
   ...
   "dependencies": {
     ...
-    "dotenv": "0.5.0"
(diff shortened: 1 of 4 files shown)


jcblw:
@maxbeatty thanks for this

jcblw:
published

PR #50, opened 2015-02-13
Add explanation env-specific variables to README

adamrneary:
The way that environment-specific files load makes perfect sense, but I found myself scanning through the tests to make sure everything behaved the way I expected. Even if the verbiage below isn't perfect, I think it may be worth mentioning within the README how environment-specific variables are loaded.


jcblw:
looks good :+1:  thanks for contrib

PR #53, opened 2015-03-01
Move to lab and use its built in test coverage.

motdotla:


motdotla:
So I'd like to move to lab instead of mocha. It gives us code coverage for free.

Otherwise, I think this is a good pull request from @maxbeatty.

Thoughts, @jessefulton, @maxbeatty, @jcblw.

.eslintrc
```diff
@@ -1,8 +0,0 @@
-{
-  "env": {
-    "node": true
```
maxbeatty: does lab set their internal eslint to node env? if not, this turns on some rules and would be worth keeping

lib/main.js
```diff
@@ -1,37 +1,38 @@
-"use strict";
+'use strict'
```
maxbeatty: [`quotes`](http://eslint.org/docs/rules/quotes.html) would be a good rule to to add to `.eslintrc`

lib/main.js
```diff
@@ -1,37 +1,38 @@
...
   /*
    * Main entry point into dotenv. Allows configuration before loading .env and .env.$NODE_ENV
-   * @param {Object} options - valid options: path (".env"), encoding ("utf8")
+   * @param {Object} options - valid options: path ('.env'), encoding ('utf8')
    * @returns {Boolean}
   */
-  config: function(options) {
-    var path = ".env",
-    encoding = "utf8";
+  config: function (options) {
```
maxbeatty: [`space-before-function-parentheses`](http://eslint.org/docs/rules/space-before-function-parentheses.html) would be a good rule to add to `.eslintrc`

lib/main.js
```diff
@@ -1,37 +1,38 @@
-"use strict";
+'use strict'
 
-var fs = require("fs");
+var fs = require('fs')
```
maxbeatty: [`semi`](http://eslint.org/docs/rules/semi.html) would be a good rule to add to `.eslintrc`

package.json
```diff
@@ -3,11 +3,8 @@
   "version": "1.0.0",
   "description": "Loads environment variables from .env file",
   "main": "lib/main.js",
-  "directories": {
-    "test": "test"
-  },
   "scripts": {
-    "test": "gulp ci"
+    "test": "lab test/* --coverage && standard"
```
maxbeatty: if we're using lab for its built-in coverage why not also use its built in linter?

jcblw: [> No configuration. Just drop it in. The easiest way to enforce consistent style in your module/project.](https://github.com/feross/standard#one-style-to-rule-them-all)

maxbeatty: what's got everyone sold on `standard`? no config?

beyond `eslint` being bundled, it also offers more rules, wider community support, and more [integrations](http://eslint.org/docs/integrations/). to use with `lab`, you just pass `-L` or `--lint`

motdotla: I like the opinionated-ness of `standard`. And aside from that the style is good.  

Particularly for an open source lib, it's one less moving part.

maxbeatty: One less dependency is one less moving part :wink: `</eslint-push>`

motdotla:
Alright, going to roll with this.

PR #55, opened 2015-03-02
Add dotenv.overload()

sadjow:
Sometimes you need to overload a system environment variable. Then
use the dotenv.overload() function.

Example:
  Your system has a HTTP_PROXY environment variable already
  defined. And your application should not use a proxy to
  communicate with another application. Then you can set your .env
  file with a variable HTTP_PROXY='' to unset this proxy.

  Or you can overload whatever environment variable you want
  to.

### DIFF
--- a/.env
+++ b/.env
@@ -16,3 +16,4 @@ EQUAL_SIGNS=equals==
 ZERO_WIDTH_CHARACTER=​user:pass@troup.mongohq.com:1004/dude
 RETAIN_INNER_QUOTES={"foo": "bar"}
 RETAIN_INNER_QUOTES_AS_STRING='{"foo": "bar"}'
+OVERLOAD_THIS_VAR='overloaded'
--- a/lib/main.js
+++ b/lib/main.js
@@ -76,11 +76,17 @@ function dotenv() {
 
       return true;
     },
-    _setEnvs: function() {
+    _setEnvs: function(overload) {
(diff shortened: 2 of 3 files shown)


maxbeatty:
why not just use non-conflicting variable names for your app by name-spacing or prefixing? allowing this sort of override could allow someone to hurt themselves by overriding a variable like `PATH`, `SHELL`, or `PWD`

sadjow:
I think that setting `PATH` inside .env file is the same as to call the application with the `PATH` variable `PATH=foo ./bin/myapp`. And I think that the name of the method `overload` is clear.

motdotla:
Let's see if anyone else weighs in on this - as wanting it.

I see this as very dangerous. Dotenv should really only be called once - anything else littered around your code is adding complexity.

@sadjow why not just instead 'overload' by setting `process.env.THAT_VARIABLE = 'something'` ?

sadjow:
Yes @motdotla . We can just use `process.env.SOME_VARIABLE = 'something'`

Thank you all for the review.

userpixel:
If someone has `PATH`, `SHELL`, or `PWD` plainly sitting in their `.env` file and chose the opt-in `override: true` as a config option, I don't think they're going to have a hard time figuring out why. Besides, the mentioned environment variable names are not something people just play around in a Node application.

PR #56, opened 2015-03-14
Standardise badges in README

matiassingers:
All using the same `?style=flat-square`.

Unfortunately the [fury.io](https://badge.fury.io/) badge didn't support flat-square, and the [shields.io](http://shields.io/) one is blue instead of green.


jcblw:
Looks good :+1:

PR #61, opened 2015-03-31
add optionality to omit error message and additionally give the optional...

### LINKED ISSUE #58: Suppress errors of missing .env file for production
Production environments normally don't have a .env file. In these cases `require('dotenv').load()` prints the following error to console:

```javascript
{ [Error: ENOENT, no such file or directory '.env'] errno: -2, code: 'ENOENT', path: '.env', syscall: 'open' }
```

Shouldn't these errors be suppressed or at least change into something less frightening?

wzrdtales:
...ity to pass the error in a callback as first parameter

Resolves #58

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -8,9 +8,11 @@ module.exports = {
    * @param {Object} options - valid options: path ('.env'), encoding ('utf8')
    * @returns {Boolean}
   */
-  config: function (options) {
+  config: function (options, callback) {
     var path = '.env'
     var encoding = 'utf8'
+
+    callback = (typeof options === 'function') ? options : callback
 
     if (options) {
       if (options.path) {
@@ -29,9 +31,20 @@ module.exports = {
         process.env[key] = process.env[key] || parsedObj[key]
       })
 
+      if (typeof callback === 'function') {
+        callback(null)
+      }
+
       return true
     } catch(e) {
-      console.error(e)
+      if (!options || !options.omitError) {
+        console.error(e)
+      }
+
+      if (typeof callback === 'function') {
+        callback(e)
+      }
+
       return false
     }
   },
--- a/test/main.js
+++ b/test/main.js
@@ -41,6 +41,17 @@ describe('dotenv', function () {
       done()
     })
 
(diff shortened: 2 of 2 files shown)


maxbeatty:
Thanks for putting this together. I know others are looking forward to eliminating the `console.log`.

My concern with making this asynchronous is that you'd have to wrap your app in the callback. I went down this path when exploring using streams. 

```js
require('dotenv').config({}, function(err) {
  // do app things
});
```

Was the intent of #58 to be able to handle the error or just to silence it?

wzrdtales:
Both.

wzrdtales:
So the PR is both, #58 only to omit it if wanted

maxbeatty:
I took a similar approach to add an option to not call `console.log` in #62

wzrdtales:
Also you don't need to wrap it, another way would be to rewrite the returns. But I really don't wanted to modify the behavior of this module right now.

```javascript

loadAsyncDependencies: function() {
   require('dotenv').config({ omitError: true }, this.continue.bind(this));
}
```

jcblw:
@wzrdtales I dont think we need a async way to handle these error since the module its self it sync. Right now you are able to handle this error doing

```javascript
var isEnvLoaded = dotenv.load()

if (!isEnvLoaded) {
  // handle env not being loaded
}
```

jcblw:
closing in favor of #62 thanks for the pr tho @wzrdtales :beers:

wzrdtales:
@jcblw No problem, one comment to the error handling. This might handle if dotenv failed or not, but it is not handling the error itself. As you can't say what exactly is the error, you just know that dotenv failed. Maybe adding a getLastError method would be a solution?

jcblw:
@wzrdtales 
I agree with you the check I posted above and having the error object is not 100% the same. The only reason that an error would be created is if there is a failure to read a file or a bug in our parser. The added benefits of having an error would be a human readable explanation of the error and a call stack  which is why, by default, this logged this out. 

I don't know if the benefits of having an error object are worth having to maintain a new method.

PR #62, opened 2015-03-31
add option to silence errors

maxbeatty:
ref #58

add option to silence `console.log` when error occurs while reading and parsing files

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -11,6 +11,7 @@ module.exports = {
   config: function (options) {
     var path = '.env'
     var encoding = 'utf8'
+    var silent = false
 
     if (options) {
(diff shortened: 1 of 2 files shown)


jcblw:
I prefer this over #61 :+1:

wzrdtales:
I don't prefer my or this solution. I'm fine with both, it would be great to get the ability to handle the error on the application that uses the module, but that's the only difference. ;)

maxbeatty:
Just thought of how you could control output based on environment in case you want the verboseness during development but not in other environments:

```js
require('dotenv').config({silent: process.env.NODE_ENV !== 'development'})
```

PR #71, opened 2015-05-20
add test that sets ENV to a string with spaces

anandkumarpatel:
looking at the test I was unsure if spaces would be included.
created a test to show that you can include spaces in the `.env` file


jcblw:
looks good, just waiting on the green light from travis. Seems like they be having some issues.

maxbeatty:
@anandkumarpatel thank you for adding this!

PR #81, opened 2015-07-23
add variable interpolation

munza:
sometimes it could be necessary to concat a process variable with another.
this commit will use the following method to do that.
```
BASIC=basic
TEST={$BASIC}test
```
will result in `{BASIC: 'basic', TEST: 'basictest'}`.


maxbeatty:
Thanks for PR but we've unfortunately already been down this road. See #42 for further discussion. Bash and JavaScript already do this _really_ well and **dotenv** just won't. It's a simple module with simple responsibilities.

PR #84, opened 2015-08-11
Add cli command

caske33:
This is a very basic cli, but could be enough for most purposes.

When installed globally, it can then be used as `dotenv <normal command>` and then the normal command will have the necessary environment variables.

Please let me know if anything needs changing. Or if this should be a seperate module.

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -1,0 +1,8 @@
+#!/usr/bin/env node
+
+var spawn = require('child_process').spawn
+var dotenv = require('dotenv')
+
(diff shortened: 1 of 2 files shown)


caske33:
I thought about it some more, and this should probably be its own module (so it can have its own dependencies). I'll fix it tonight.

caske33:
See https://github.com/caske33/dotenv-cli and https://www.npmjs.com/package/dotenv-cli

PR #85, opened 2015-08-14
adhere to padded-blocks

LinusU:
We are in the process of adding back the rule `padded-blocks` into `standard`. It was temporarily disabled because of a bug in eslint which now has been fixed.

This pull request makes sure that your coding style is compliant with the new version.

Please see feross/standard#170 for more info.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     "semver": "^4.3.6",
     "should": "4.4.2",
     "sinon": "1.12.2",
-    "standard": "^2.10.0"
+    "standard": "^5.1.0"
   },
   "dependencies": {}
 }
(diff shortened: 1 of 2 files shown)


maxbeatty:
Thanks for taking this initiative, @LinusU! We're currently using [standard@2](https://github.com/motdotla/dotenv/blob/master/package.json#L31) and it looks like the current major version is 5

```bash
$ npm info standard | grep version:
  version: '5.1.0',
```

Will this change to `padded-blocks` be a major version bump? If so, perhaps this PR should include updating the dep version when it's ready

LinusU:
No problem :rocket: 

The plan is to release it as a minor version, most probably 5.2. I'll bump it to `^5.1.0`, that's what I tested against (+ plus the padding rule). Sounds good?

LinusU:
Should be good to go :+1:

PR #87, opened 2015-09-06
add returns option to return as object

flrngel:
This commit is for users who wants `.env` environments as Object, not extends `process.env`


flrngel:
close to re-pr due to guide line

PR #88, opened 2015-09-06
add returns option to return as object

flrngel:
This commit is for users who wants `.env` environments as Object, not extends `process.env`

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 <img src="https://raw.githubusercontent.com/motdotla/dotenv/master/dotenv.png" alt="dotenv" align="right" />
 
-Dotenv loads environment variables from `.env` into `ENV` (process.env).
+Dotenv loads environment variables from `.env` into `ENV` (process.env for default).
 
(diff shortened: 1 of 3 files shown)


maxbeatty:
why not use [`parse`](https://github.com/motdotla/dotenv#parse) directly? we expose and document it for this purpose

flrngel:
@maxbeatty thank you for comment. As you know, `parse` uses string or buffer as parameter.
Way to use `.env` file without using writing `file reading code` outside library, `config` function was only option.

maxbeatty:
`parse` accepts a string or buffer because that's what's returned from [fs.readFileSync](https://nodejs.org/api/fs.html#fs_fs_readfilesync_filename_options). Give that your path to your `.env` file and you should be all set.

```javascript
var fs = require('fs')
var dotenv = require('dotenv')

var parsedObj = dotenv.parse(fs.readFileSync('.env'))
```

Thanks for the idea but since this is less code than adding another configuration option, I'm going to opt to keep things as-is

Issue #89, opened 2015-09-10
Importing dotenv in ES6


Sparragus:
Support on ES6 has been tricky for me. Yesterday I tried preloading dotenv with `iojs -r dotenv/config index.js`. However, it didn't work and I couldn't understand why. My app did nothing. After running the app, a second or two later it would finish without doing anything.

I ended up with a very simple way to use dotenv. I simply import it like this: 
```es6
import {} from 'dotenv/config'
import somethingElse from 'somethingElse'
...
[the rest of your code]
```

This works because of how ES6 modules imports modules. Before doing anything else in a file (say index.js) it does all the imports first. It does a depth first traversal of these imports, executing any code inside it. If we import dotenv first, it will execute config and add any env variables before doing anything else in the code. 

From [ES6 In Depth: Modules](https://hacks.mozilla.org/2015/08/es6-in-depth-modules/):
> When you run a module containing an import declaration, the modules it imports are loaded first, then each module body is executed in a depth-first traversal of the dependency graph, avoiding cycles by skipping anything already executed.

So, importing dotenv on the first line of a bootstrap file in an app will set the env vars for anything that might use them.

I suggest we add a brief section about importing dotenv with ES6 modules to the README. But before sending a pull request and wanted to open it up for discussion here.

Any thoughts?

motdotla:
@jcblw you've done some work with ES6 right?

maxbeatty:
@Sparragus can you provide a little more code so we can try to reproduce? I've added some examples of how to use dotenv in various ways in #90 including ES6 preloading. Are you using babel or something else for your `import` syntax? I don't believe the latest node v4.0.0 supports that out of the box yet.

I would caution against using `dotenv/config` in any context but preloading because it'll be doing additional work parsing command line arguments. `import dotenv from 'dotenv'` should do the trick

jcblw:
I believe since we are not exporting out a *key* for` dotenv` the default import behavior of `babel` @maxbeatty mentions should work.

```javascript
import dotenv from 'dotenv'
dotenv.config()
```

I would also be interested in how your running into an issue with preloading the module. Can you post the version of node/iojs and dotenv that you are using?

Sparragus:
@maxbeatty and @jcblw 

I've uploaded a [repo with sample code](https://github.com/Sparragus/dotenv-es6-debugging) showing everything I mentioned on my first message. 

[...]

PR #90, opened 2015-09-10
add examples for how to use module in different ways

maxbeatty:
to help w/ #89, I've put together a few examples for how to use dotenv in different ways. in the future if someone has a problem, hopefully we can reduce it to its simplest form using one of the examples

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,0 +1,6 @@
+coverage/
+examples/
+test/
+.editorconfig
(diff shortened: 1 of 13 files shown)


motdotla:
This is great. Let's give some time for a response to this: https://github.com/motdotla/dotenv/issues/89#issuecomment-139353056

But otherwise. :+1:

PR #91, opened 2015-09-16
standard

feross:
We use this package in our test suite as a sanity check and the rules have changed a bit. :+1:

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -34,7 +34,7 @@ module.exports = {
       })
 
       return true
-    } catch(e) {
+    } catch (e) {
       if (!silent) {
         console.error(e)
       }


jcblw:
:+1: sweet thanks man!

maxbeatty:
why wasn't travis failing before this change?

feross:
Because I just published the new version a few minutes ago and you hadn't re-run the tests :)

maxbeatty:
thanks for being proactive but this could be avoided by following semver

feross:
I'm a huge semver proponent, but it's not that simple. For a linter, any change, even in dependencies, is likely to break someone's code.

For example, eslint publishes patch versions when they improve a rule to detect some case that was being missed before. Makes sense to be a patch, since it's a bugfix: the rule wasn't working correctly before. But to end users, that patch version could cause eslint to start failing on code that passed before. This isn't theoretical. This happens all the time.

So, to be safe, literally every time we bump a dep, we would need to publish a major version. This isn't realistic or even useful. Regarding this PR, it was always the case that you should have spaces before/after keywords - it just wasn't catching this case.

PR #93, opened 2015-09-17
Bumping dependencies

motdotla:


jcblw:
:ok_hand: looks good

motdotla:
@maxbeatty your thoughts?

And @jcblw @maxbeatty shall I version this up 1 patch to: `1.2.1` ?

jcblw:
I don't think we need to publish this seems like a development heavy amount of changes. I wouldn't be opposed to making a bump in the version or a tag to track the development changes.

.travis.yml
```diff
@@ -1,6 +1,6 @@
 language: node_js
 
 node_js:
-  - iojs
   - 0.12
-  - 0.10
+  - iojs
```
maxbeatty: perhaps we want to follow [node's LTS](https://github.com/nodejs/LTS) for our matrix. today that would be 0.10, 0.12, and 4

package.json
```diff
@@ -26,11 +26,11 @@
...
   "devDependencies": {
     "babel": "^5.8.23",
-    "lab": "^5.3.0",
-    "semver": "^4.3.6",
-    "should": "4.4.2",
-    "sinon": "1.12.2",
-    "standard": "^5.1.0"
+    "lab": "^5.17.0",
+    "semver": "^5.0.3",
+    "should": "7.1.0",
```
maxbeatty: any reason why some deps use exact versions and others use caret?

maxbeatty:
agree w/ @jcblw that we don't need to make a release since what module users install won't change. having zero dependencies has its perks :ice_cream:

.travis.yml
jcblw: was talking to @motdotla about `0.10` earlie; the update to standard is breaking our `0.10` install/test I am down to ditch the `iojs` tho.

package.json
motdotla: No good reason. I'll change to carets

.travis.yml
motdotla: Yeah, it literally was failing on `npm install`. Trying again though.

motdotla: It just passed. Going to try again up to travis

PR #95, opened 2015-09-17
use standard eslint config

maxbeatty:
attempt to work around https://github.com/feross/standard/issues/261 by using their eslint config and the built-in linter of lab

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,0 +1,3 @@
+{
+  "extends": "standard"
+}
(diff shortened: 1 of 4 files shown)


maxbeatty:
[this github dependency](https://github.com/Flet/standard-engine/commit/6f1bc46f35c1ab826cdeaf8b94887dc6d22c1ea4) in a sub-dependency of standard is what caused the node v10 issues. the eslint config and plugin I've added here (which are also inside of standard) do not have any dependencies so using them should prevent these kinds of failures in the future.

@motdotla @jcblw thoughts on using `.eslintrc` instead of `standard` bin?

motdotla:
Root problem was resolved. Let's stick with `standard` bin. Changes here: https://github.com/motdotla/dotenv/pull/93

Issue #77, opened 2015-06-29
Support for Arrays?


decoursin:
Hi, is their support for arrays? If so, what's the syntax like? Thank you.

jcblw:
We can not set any thing other then strings to the `process.env` object. Those get type casted on the setter. Eg.

```shell
node -e "process.env.ALPHA=['a', 'b']; console.log(typeof process.env.ALPHA);"
# will be "string"
```

but you can do this.

```shell
# .env file
ALPHA=a,b,c,d
```

```javascript
require('dotenv').config();
var alpha = process.env.ARRAY_ALPHA.split(',');
``` 

Sorry.

decoursin:
Sounds good thank you so much that might work well enough

niftylettuce:
Hey guys I had this same issue so I built this plugin for `dotenv` at https://github.com/niftylettuce/dotenv-parse-variables.  Hope it helps 👍

andywer:
Why not add support for a generic solution like this? Achieves the same, but in an explicit and non-breaking way.

```ini
# .env
HOSTS[]=localhost
HOSTS[]=example.com
```

mckenzieja:
@andywer 

Has your implementation been tested working? I am considering your method.

andywer:
Haven‘t actually implemented it. A simple workaround that doesn‘t require anything new: Put JSON in the variable and `JSON.parse()` or make it a comma-separated list of items that you can easily `.split(',')` 😉

drummy2:
> Haven‘t actually implemented it. A simple workaround that doesn‘t require anything new: Put JSON in the variable and `JSON.parse()` or make it a comma-separated list of items that you can easily `.split(',')`

I think this then means you have to have it all on a single line?

appsparkler:
Hello People, 

I wanted `JSON`.  This is what worked for me:

```
// .env.development
...
WEBPACK_HTML_PLUGIN={"template": "src/index.html", "filename":}
SOME_ARRAY=["hello", "world"]
...


// module-file.js (accessing the .env variable)
typeof JSON.parse(process.env.WEBPACK_HTML_PLUGIN); // object
Array.isArray(JSON.parse(process.env.SOME_ARRAY)) // true
```

Good Luck...

PR #96, opened 2015-09-23
Retain parsed variable types.

aichholzer:
The fact that all `.env` variables are parsed as `strings` was annoying me a bit since one would expect that all variables loaded through it are what they should be. In my case, for example, I had something like `LOAD_PHOTOS=false` and then I was trying to use the loaded variable as:

```
var loadPhotos = process.env['LOAD_PHOTOS'];
if (!loadPhotos) {
    ...
}
```

which of course never worked, since `loadPhotos = 'false'` and therefore never `!loadPhotos`.

This pull request addresses this issue and variables retain their type. Useful, instead of having to write one's own casting logic after `.env` is done loading.


maxbeatty:
[`process.env`](https://nodejs.org/api/process.html#process_process_env) follows [environ(7)](http://man7.org/linux/man-pages/man7/environ.7.html) which is described as being "an array of pointers to strings". I take that as a very intentional decision to have all values as strings so it's always clear what you'll get.

`isNaN` has [some quirks](https://github.com/lodash/lodash/blob/master/lodash.core.js#L2736-L2765). What if a username or password is completely numeric and someone expects to use it as a string? With this addition, they'd have to undo this work.

As far as booleans are concerned, there are two options that'll work today without any changes.

1. Omit the value from your .env file. This would cause `process.env['LOAD_PHOTOS']` to be `undefined` and `!loadPhotos` to evaluate to `true`.
2. Construct a `Boolean` (or other type) with `process.env['LOAD_PHOTOS']` as your initial value. This would make your own code more explicit as to what type `loadPhotos` should be.

.env is a plain text file that loads strings into `process.env` and that simplicity brings consistency for everyone who uses it.

jcblw:
This would add the feature only to the parser but would not effect the main use of `dotenv` since you cannot set a variable with a type different then a string to process.env. #77

screen shot 2015-09-23 at 11 43 21 am

Since this is not going to be used by a majority of the users of `dotenv`  I would say this would be a great opportunity to make another module to handle this case rather then add it to the core of `dotenv`. Something like...

```javascript
const cast = require('smart-cast-obj')
const env = cast(dotenv.parse('IS_TRUE=true')); // {IS_TRUE: true}
```

If there is a need for this tho in core I would like to see an option to turn this on.

maxbeatty:
:+1: piping parse output to another module for type casting

aichholzer:
> "an array of pointers to strings". ... it's always clear what you'll get.

I agree, however, this is exactly the point; a user would assume that the parser is smart enough to return the value as is. Affecting the behaviour of `process.env` for it to accept anything other than a string is not the case here. This also applies to your reply @jcblw; this is a mere fix on the parser.

The simplest solution would be to stick with what I was doing so far; `export` all variables and then just use plain `process.env` to get them and applying some cast logic -if needed- for each one of them. Switching to `.env` was supposed to remove the whole `export` process, and since it's a parser -per se- one could expect it to handle some logic, hence my proposal.

> Since this is not going to be used by a majority of the users of `dotenv`

Quite an absolute statement @jcblw - I would not be so sure about that.

jcblw:
from the inspiration for this project
> [Shim to load environment variables from .env into ENV in development.](https://github.com/bkeepers/dotenv/blob/master/README.md)

the second line of our readme
> Dotenv loads environment variables from .env into ENV (process.env).

I will give you that @aichholzer; it is an absolute statement, but I still believe it should not be on by default.

Is there any issues to segmenting the casting to after the variable has been parse (line break, variable stuff). From a quick look it doesn't seem like it would hurt... as well as it would enable parsing to happen on possible variables that are already inside the ENV.

aichholzer:
Having that said, I second the idea of being able to enable this feature, assuming the developer knows what he is doing -and what he is loading- it would be safe to provide an `on` || `off` feature. Perhaps something like this:

```
require('dotenv').load({ 'preserveType': true });
```

or

```
require('dotenv').load({
  preserveType: [
    'LOAD_PHOTOS',
    'VALUE_OF_PI'
  ]
});

// {boolean} LOAD_PHOTOS=true
// {number} VALUE_OF_PI=3.1416
```

sort of explicitly declaring for which values to retain the type.

maxbeatty:
"retain" has been used a few times in this thread and it is misleading. `.env` files are plain text (no types). The parser's job is to identify keys and values and assign them to `process.env` if a value doesn't already exist (with a few other subtleties relating to strings). At that point, `process.env` is just an object that you can manipulate. That's a great point to hand off responsibility to the programmer or another module to alter it as they like (e.g. casting types). This would keep `dotenv` focused and allow another module to expand where it might not make sense for this module (i.e. arrays, objects, sets, and other types).

aichholzer:
`retain` is indeed the wrong term to be used, just being used in spite of a more appropriate one. We know what is going on (and what `retain` refers to, hopefully) and whoever reads this request will understand what is going on as well.

Furthermore, I guess we are also well acquainted with the fact that `environment variables` are `strings`, which is not the subject at hand.

This is not meant to be a religious discussion about `environment variables` being strings and only strings, we know that. It's about proving a solution that fixes something that, IMHO, seems cumbersome.

motdotla:
Good thoughts here on both ends, and thanks for the PR @aichholzer. But I'm with @maxbeatty and @jcblw:

:+1: piping parse output to another module for type casting rather than this living in dotenv.

Would love to see you create this as a separate repo @aichholzer (maybe in a way @jcblw suggested).

```
const cast = require('smart-cast-obj')
const env = cast(dotenv.parse('IS_TRUE=true')); // {IS_TRUE: true}
```

aichholzer:
@motdotla will think of something along those lines. `;)`

bjoshuanoah:
I like the idea of extending dotenv with an external module:

    var dotenv = require('dotenv'),
        cast = require('smart-cast-obj')(dotenv);
    dotenv.load().cast();

Something like that, where you can add new methods. 

This would probably require a change in the api to allow for intercepting the load/parse methods

maxbeatty:
Another benefit of dotenv's simple, completely exposed API (read: JavaScript Object) is that you can wrap and manipulate it however you want. Take this simple example:

```js
dotenv = { 
  load: function () { 
    console.log('loaded')
  }
}

cast = function (dotenv) { 
  var prevLoad = dotenv.load
  var myCoolThing = { 
    cast: function () { 
      console.log('casted')
    }
  }
  dotenv.load = function() { 
    prevLoad()
    return myCoolThing
  }
}

cast(dotenv)

dotenv.load().cast()
// => loaded, casted
```

niftylettuce:
Hey guys I had this same issue so I built this plugin for `dotenv` at https://github.com/niftylettuce/dotenv-parse-variables.  Hope it helps 👍

aichholzer:
👍

PR #98, opened 2015-10-01
add variable expansion test and faq

maxbeatty:
this is a common request and a main point of resistance for me is dealing with the possibility that existing API keys or other random strings could be treated as variables and expanded incorrectly. hopefully this saves people time digging through old issues and helps build a better test suite to support the new feature.


motdotla:
:+1:

PR #105, opened 2015-11-01
Change comment in lib/main.js - remove .env. reference

sudodoki:


jcblw:
:+1:

maxbeatty:
> Allows configuration before loading .env

Still doesn't make sense because this method does load .env

motdotla:
It is a little fuzzy but that sentence has been there for a while and this at least adds more clarity. We can create a separate PR to make a clearer sentence.

Gonna merge in. :+1:

PR #118, opened 2016-01-08
Parse exported variables.

### LINKED ISSUE #112: Ignore leading `export ` for direnv.
It may be nice to add support that ignores leading `export\s+` on each line, so that we can interoperate with [direnv](https://github.com/direnv/direnv) files.

adjohnson916:
Fixes #112.

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -53,7 +53,7 @@ module.exports = {
     // convert Buffers before splitting into lines and processing
     src.toString().split('\n').forEach(function (line) {
       // matching "KEY' and 'VAL' in 'KEY=VAL'
-      var keyValueArr = line.match(/^\s*([\w\.\-]+)\s*=\s*(.*)?\s*$/)
+      var keyValueArr = line.match(/^\s*(?:export)?\s*([\w\.\-]+)\s*=\s*(.*)?\s*$/)
       // matched?
       if (keyValueArr != null) {
         var key = keyValueArr[1]
--- a/lib/main.js
+++ b/lib/main.js
@@ -53,7 +53,7 @@ module.exports = {
     // convert Buffers before splitting into lines and processing
     src.toString().split('\n').forEach(function (line) {
       // matching "KEY' and 'VAL' in 'KEY=VAL'
-      var keyValueArr = line.match(/^\s*(?:export)?\s*([\w\.\-]+)\s*=\s*(.*)?\s*$/)
+      var keyValueArr = line.match(/^\s*(?:export\s+)?([\w\.\-]+)\s*=\s*(.*)?\s*$/)
       // matched?
       if (keyValueArr != null) {
         var key = keyValueArr[1]
--- a/test/.env
+++ b/test/.env
@@ -21,3 +21,4 @@ ALWAYS_VARIABLE=a0${beep}${b00p}
 export EXPORTED=exported
 EXPORT=export
 export=export
+exportedKey=foo
--- a/test/.env
+++ b/test/.env
@@ -19,3 +19,5 @@ RETAIN_INNER_QUOTES_AS_STRING='{"foo": "bar"}'
 INCLUDE_SPACE=some spaced out string
 ALWAYS_VARIABLE=a0${beep}${b00p}
 export EXPORTED=exported
+EXPORT=export
+export=export
--- a/test/.env
+++ b/test/.env
@@ -18,3 +18,4 @@ RETAIN_INNER_QUOTES={"foo": "bar"}
 RETAIN_INNER_QUOTES_AS_STRING='{"foo": "bar"}'
 INCLUDE_SPACE=some spaced out string
 ALWAYS_VARIABLE=a0${beep}${b00p}
+export EXPORTED=exported
--- a/test/main.js
+++ b/test/main.js
@@ -215,5 +215,10 @@ describe('dotenv', function () {
       parsed.export.should.eql('export')
       done()
     })
+
+    it('not confused by variable name starting with export', function (done) {
+      parsed.exportedKey.should.eql('foo')
+      done()
+    })
   })
 })
--- a/test/main.js
+++ b/test/main.js
@@ -204,5 +204,10 @@ describe('dotenv', function () {
       parsed.ALWAYS_VARIABLE.should.eql('a0${beep}${b00p}')
       done()
     })
+
+    it('ignores leading export', function (done) {
+      parsed.EXPORTED.should.eql('exported')
+      done()
+    })
   })
 })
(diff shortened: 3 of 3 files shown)


maxbeatty:
Will you add documentation about why and how you would use this new feature? Another good test to add would be to ensure that `export` can still be used as a normal variable (`EXPORT=true`, `export=true`, etc).

I haven't used [direnv](http://direnv.net/) before but after reading their homepage, it's unclear to me why you would use their `.envrc` file with dotenv. If `direnv` loads `.envrc` and exports any defined variables to the shell, then any node process started in that shell should already have those variables assigned to `process.env` before dotenv comes into play. Here's a quick proof of concept:

```bash
$ export FOO=bar
$ echo $FOO
bar
$ node
> process.env.hasOwnProperty('FOO')
true
> process.env.FOO
'bar'
```

adjohnson916:
@maxbeatty Some of mine and my colleagues' shells don't have direnv. On shells with direnv, I want to use a direnv config rather than a dotenv config, so as to expose the variables to other tools. But on shells without direnv, I'd like at least dotenv to work. I'd like to use the same direnv  config file in both cases to avoid dual maintainence.

If you're open, I can add a test for "export" as a variable name. This change shouldn't break anything. And I can add documentation - where to, the README?

maxbeatty:
I appreciate you wanting to make other developers' lives easier. I'm not :100: about trying to support other file formats given the scenario you've outlined. Part of the motivation for this module is convention, and in this instance, that means using a consistent file format. I'm going to leave this open so others can weigh in. The more robust the tests and documentation are around this addition, the better.

adjohnson916:
@maxbeatty Okay. So you're saying that currently this module's scope is limited to providing environment variables only to Node scripts and specifically only those that explicitly use this module. So for example if I have some npm scripts that run external tools that rely on the same environment variables as my app that I already have configured in my `.env`, I have to repeat my configuration for those variables elsewhere (e.g. using direnv, an `.npmrc` file, etc.).

With this change, a developer without direnv could easily run `source .env ` to export those environment variables for other tools. This could prefix an npm script as well - e.g. `"test": "source .env; mocha"`.

I'd like to see this module be able to adapt to what I think could become a pretty common use case.

For now, to accomplish that, another tool is required to pull the variables out of the `.env` file and export them. Here's a tiny Bash script I've written as basically a more portable direnv substitute - I can put this in each project as `exenv.sh` and expect others to run `./exenv.sh` each time they start working:
```
# export environment variables from a file like Node.js's dotenv
[ -f ./.env ] && while read LINE; do
  if [[ ! $LINE =~ [^[:space:]] ]] ; then ; continue; fi
  LINE=`echo $LINE | sed -Ee 's/[ \t]*export[ \t]+//g'`
  export $LINE
done < ./.env
```

adjohnson916:
Added a test to confirm that despite this change we still support `export` and `EXPORT` as variable names.

maxbeatty:
Correct. This is not a general purpose utility to populate environment variables from a file. It is intended to help developers configure their Node.js applications based on the environment (e.g. database passwords, API tokens, hostnames).

Could you call that bash script using direnv? Every developer on your team could follow the convention of using a `.env` file, and those that want to use direnv would be able to opt in to loading all of the app variables into their shell. This would also allow dotenv to stick to the current informal configuration file standard.

electrickite:
I'd like to see this feature merged as well. In addition to adding flexibility to the `.env` file, this also would provide consistency with similar projects in other languages. The popular implementations in Ruby and PHP both allow the `export VAR=value` syntax.

https://github.com/bkeepers/dotenv
https://github.com/vlucas/phpdotenv

adjohnson916:
Thanks for the support @electrickite.

lib/main.js
```diff
@@ -53,7 +53,7 @@ module.exports = {
     // convert Buffers before splitting into lines and processing
     src.toString().split('\n').forEach(function (line) {
       // matching "KEY' and 'VAL' in 'KEY=VAL'
-      var keyValueArr = line.match(/^\s*([\w\.\-]+)\s*=\s*(.*)?\s*$/)
+      var keyValueArr = line.match(/^\s*(?:export)?\s*([\w\.\-]+)\s*=\s*(.*)?\s*$/)
```
jcready: The non-capture group needs to include the `\s` and you need to replace the `*` after it with `+`:

```js
var keyValueArr = line.match(/^\s*(?:export\s+)?([\w\.\-]+)\s*=\s*(.*)?\s*$/)
```

Otherwise you'll end up matching `exportedKey=foo` as `['exportedKey=foo', 'edKey', 'foo']`

maxbeatty: Looks like a good test case to add

adjohnson916: @jcready @maxbeatty Oh duh, I can definitely make this fix and add tests very soon... please stand by.

adjohnson916:
@jcready @maxbeatty I've pushed a fix for variable names starting with 'export'. Let me know if you want me to rebase and force push at any point to clean up the history. Hope you're more open to merging this in - please let me know any other concerns preventing that from happening!

adjohnson916:
Looks like there's been some support for this (a commenter and 3 votes for my latest fix), but I don't see any motion on getting it merged.

For now, I've resorted to using a little CLI tool I just wrote called [dotenv-export](https://github.com/adjohnson916/dotenv-export) that lets you source & export `.env` variables into your shell.

maxbeatty:
👍 a focused module for this behavior is a better way to go long term

jcready:
I'm confused. What was the reason for closing this PR, @maxbeatty?

maxbeatty:
@jcready [dotenv-export](https://github.com/adjohnson916/dotenv-export) does what this PR was proposing to do. Those who want that functionality can have it and maintainers of this module don't have to maintain that functionality. Win-win in my book

bethesque:
I was disappointed to find it didn't support the "export X=Y" syntax as the Ruby one does. dotenv-export does not provide the same functionality os dotenv. I would very much like to see this feature supported.

a666:
I'm hoping this will be reopened, I see very few good reasons to not have this functionality in the reference javascript dotenv implementation.

Literally every other dotenv (ruby: https://github.com/bkeepers/dotenv , python: https://github.com/theskumar/python-dotenv , go: https://github.com/joho/godotenv , rust: https://github.com/dotenv-rs/dotenv , elixir: https://github.com/avdi/dotenv_elixir) supports having a optional `export` if you have the need to source your `.env` file.

I was in a recent vuejs project, and the team was utterly staggered by the fact that the environment variables where not being read from a standard `.env` file, until, you know, we realized that `export` was not parseable in a dotenv file...

Issue #107, opened 2015-11-18
Option to overwrite env variables


tmarciszewski:
Hi,

I'm wondering why optional overwriting of existing env variables is not supported at the moment. Are there any problems with that feature?

It was implemented here: https://github.com/jordansexton/dotenv

My use case is as follows:

I want to set env variables in dev environment. I'm using gulp to restart project if some files change. In my case environment variables should change if I change .env file.

Should I prepare pull request with this feature?

Best Regards
Tomasz

maxbeatty:
Can you provide a simplified example of your setup? If your gulp restart isn't restarting the process, then `process.env` will already be populated and [not reassigned](https://github.com/motdotla/dotenv/blob/master/lib/main.js#L33).

Issue #102, opened 2015-10-23
Document behaviour when environment variable is already set on the machine


silhouettes:
My team had multiple issues between environments this week that, after much confusion, boiled down to environment variables not being set by dotenv because they were already set to different values on certain machines.

Of course we were being silly and this little detail of dotenv's behavior flew over our heads, but I think it would be extremely helpful to document this behavior as it is not mentioned anywhere and the (possibly erroneous) temptation is to assume that the environment of the application is whatever is in the .env file.

Also, maybe a boolean flag to specify which thing 'wins' would be very helpful, although not essential!

geof90:
:+1:

maxbeatty:
Sorry for the trouble and confusion! [Here's the implementation if you're still curious](https://github.com/motdotla/dotenv/blob/master/lib/main.js#L33). We should definitely document this.

I'm waiting for feedback on #101 and could include it there. Even if that doesn't get accepted, a PR with ideas on what and how to explain the behavior would be helpful.

silhouettes:
Okay, no worries and thanks! I'll submit a quick PR with an update on the FAQ section, and you can feel free to modify it as desired.

Issue #121, opened 2016-01-21
CLI doesn't appear to load env


marr:
I have a simple example, my .env file includes: `DEBUG=app:test`.
I run a test script thusly: `node -r babel-register -r dotenv/config bin/test.js`
The contents of the script are:
```
import _debug from 'debug'
const debug = _debug('app:test')
debug(process.env.DEBUG)
```
Nothing is output from the CLI.
However, if I run the script with `DEBUG=* node -r babel-register -r dotenv/config bin/test.js` I see the output:`app:test * +0ms`

maxbeatty:
This smells like #114 where babel is compiling the source which messes with load order. I verified [the es6-preload example](https://github.com/motdotla/dotenv/tree/master/examples/es6-preload) still works. You might want to `console.log(process.env.DEBUG)` to see if it's populated. If it is, `debug` is getting called too early in the babel compiling.

marr:
Ok, the console.log does output the correct value. However I initially thought there was a problem because I ran `node -r dotenv/config` and then ran `process.env.DEBUG` in the repl. `undefined` is output.

I guess thats just an oddity with how the repl runs?

hyperh:
I'm having the same issue. I do get the correct output when I do `console.log(process.env.DEBUG)`, but `debug` still not working for me.

Issue #115, opened 2015-12-29
add 'replace' option to overwrite existing process.env keys


guymguym:
hi
the code ignores existing keys if already found in the process.env and does not overwrite them.
for some cases that works fine, however in some cases we keep a .env file which is expected to overwrite existing environment variables even if already present.
so I think an option would be helpful like that - 
```load({replace: true})``` or ```load({force: true})```
what do you think? 
would love to submit a PR if that makes sense.
thanks!

**Comment by maxbeatty** [comment]
Can you expand on your use case for needing to overwrite existing environment variables?

The current logic is based on the assumption that you have control over the environment you're running your code in meaning you should be able to control the order in which your environment variables are assigned. You can load multiple files to emulate overwriting.

*.env.override*
```
NODE_ENV=override
```

*.env*
```
NODE_ENV=example
```

*index.js*
```js
const assert = require('assert')
const dotenv = require('dotenv')
dotenv.load({ path: '.env.override' })
dotenv.load()
assert.equal(process.env.NODE_ENV, 'override')
```

If `.env.override` is missing in some environments, it'll safely fail and move on.

guymguym:
Hi @maxbeatty Thanks for the fast reply.

In my case the environment is coming from the parent process (currently it is supervisord), and I simply want to override any inherited environment with my `.env`. We actually encountered a case when the application env got merged into parent and later was not able to be modified using `.env`. 

So the case is that the `.env` file represents an overriding environment over the parent's env, and I also want it to be used regardless of the parent that spawned it.

Anyhow, my suggestion is indeed a tiny addition, that may or may not be useful to anyone else, but it's not very disturbing, right?

Many thanks!
Guy

guymguym:
and according to your terminology, this option is best described as `load({override: true})` which works well for me.

maxbeatty:
Sorry, I'm still not totally understanding the use case. I get you want `process.env[key] = val || process.env[key]` more or less, but it would be helpful to understand what exactly you're trying to override. Is it a database password? API key? Feature flag? Something else entirely?

Are there certain environments where this is more useful than others? Testing environments like CI can sometimes be hard to configure thus needing more options. 

Some quick pseudo code or an example repo would really help :smile_cat:

guymguym:
one specific case is enabling/disabling a specific service that we run internally.
another is listening ports.
but these are just examples.
the basic thing is that we have a `.env` that needs to override the parent env.
makes sense?

[...]

PR #123, opened 2016-01-25
Add option to overwrite existing process.env variables with .env values

galenandrew:
**This pull request adds a `force` option that (when set) will write over existing environment variables with values from the application's `.env` file.** It is disabled by default so that it is non-breaking.

Also, this pull request fixes a false-positive test due to case-sensitive `process.env` variables.

---

Here is some background on the proposed option. I'm open to discussing alternatives.

Because this module only applies the `.env` values to the application scope (i.e. `process.env`), overwriting existing values will not affect other programs outside of the application that rely on machine-specific variables (i.e. non-destructive to the machine's environment).

When enabled, it provides more consistent and predictable behavior to the application's environment – in other words the values you see in the `.env` file are the values your application will get (think _WYSIWYG_).

_Related to #102, #107, #115_

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,6 +102,17 @@ using this option.
 require('dotenv').config({encoding: 'base64'});
 ```
 
+#### Force
+
+Default: `false`
+
+By default, dotenv **will not** write over existing environment variables in `process.env`.
+You can change this behavior by setting this option to `true`.
+
+```js
+require('dotenv').config({force: true});
+```
+
 ## Parse
 
 The engine which parses the contents of your file containing environment
@@ -151,7 +162,7 @@ No. We **strongly** recommend against having a "main" `.env` file and an "enviro
 
 ### What happens to environment variables that were already set?
 
-We will never modify any environment variables that have already been set. In particular, if there is a variable in your `.env` file which collides with one that already exists in your environment, then that variable will be skipped. This behavior allows you to override all `.env` configurations with a machine-specific environment, although it is not recommended.
+By default, this module will never modify any environment variables that have already been set. In particular, if there is a variable in your `.env` file which collides with one that already exists in your environment, then that variable will be skipped unless using the [force option](#force).
 
 ### Can I customize/write plugins for dotenv?
 
--- a/lib/main.js
+++ b/lib/main.js
@@ -12,6 +12,7 @@ module.exports = {
     var path = '.env'
     var encoding = 'utf8'
     var silent = false
+    var force = false
 
     if (options) {
       if (options.silent) {
@@ -23,6 +24,9 @@ module.exports = {
       if (options.encoding) {
         encoding = options.encoding
       }
+      if (options.force) {
+        force = options.force
+      }
     }
 
     try {
@@ -30,7 +34,7 @@ module.exports = {
       var parsedObj = this.parse(fs.readFileSync(path, { encoding: encoding }))
 
       Object.keys(parsedObj).forEach(function (key) {
-        process.env[key] = process.env[key] || parsedObj[key]
+        process.env[key] = force === true ? parsedObj[key] : process.env[key] || parsedObj[key]
       })
 
       return parsedObj
--- a/test/main.js
+++ b/test/main.js
@@ -65,12 +65,30 @@ describe('dotenv', function () {
       done()
     })
 
-    it('does not write over keys already in process.env', function (done) {
+    it('does not write over keys already in process.env by default', function (done) {
       process.env.TEST = 'test'
       // 'val' returned as value in `beforeEach`. should keep this 'test'
       dotenv.config()
 
       process.env.TEST.should.eql('test')
+      done()
+    })
+
+    it('writes over keys already in process.env when force option is enabled', function (done) {
+      process.env.test = 'test'
+      // 'val' returned as value in `beforeEach`
+      dotenv.config({ force: true })
+
+      process.env.test.should.eql('val')
+      done()
+    })
+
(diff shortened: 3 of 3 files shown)


maxbeatty:
> Also, this pull request fixes a false-positive test due to case-sensitive process.env variables

Please open a separate issue as your changes here don't clearly identify the problem or solution.

IRT wanting to override `process.env` with `.env`, the only relevant linked issue is the contentious #115 where no good use case was presented. Why do you need this feature? What root problem are you trying to solve?

I think you're misinterpreting VI and X from 12 Factor. Nine is about sharing state across processes which is unrelated to environment configuration (i.e. don't stash stuff in `process.env` to access later). Ten is about backing services and less about "parity" of environments (i.e. use MySQL during development and production, not SQLite locally and then MySQL elsewhere). By using `.env` in development and production (or other environments), you're creating parity of configuration management.

I am concerned that with this new option people will inadvertently undo environment parity that node provides (i.e. PATH, HOME, etc.), become frustrated, and open issues here (increasing support costs for maintainers). #121 is an example of thinking this module was at fault when in reality it was user error. Minimizing the ways this module can be used minimizes the ways it can be _misused_. Understanding _why_ this is needed is very important to consider adding more configuration. Good reasons may exist (I hope they do).

galenandrew:
@maxbeatty, thanks for the quick response! I apologize mine was a bit delayed and lengthy.

Regarding the "false-positive test", you can see my relevant change in this commit – 6c26b6a. Including it in this pull request does confuse things a bit so I'll remove it and open a new one.

Also, after re-reading VI and X from 12 Factor you're correct in that they don't really apply here. I do however disagree that #102 and #107 are not related. In #102 @silhouettes mentions _"maybe a boolean flag to specify which thing 'wins' would be very helpful"_ and #107 is titled _"Option to overwrite env variables"_ (the specific use case mentioned was gulp-related, but the feature requested is the same).

As for you concern that this new option will cause confusion…
- you already have people opening issues as a result of being confused why their `.env` file values are not taking affect
- clear documentation reduces confusion and misuse more than minimizing functionality (this PR includes clear documentation)
- the default option does not change current module behavior so the only time it will change is when a developer explicitly sets the option…so they will know what it does

---

Regarding the main discussion, there definitely seems to be some contention around this request in other issues and I'm not sure why – seems to be a holy war of sorts. I can understand reasoning for both sides, but feel it should be a implementation choice (option) vs a module-enforced restriction.

I've also noticed a [shift in the maintainers preference](https://github.com/motdotla/dotenv/pull/101#issuecomment-154740515) with the introduction of returning the parsed object in `dotenv@2.x.x` – having users of dotenv do things outside of the module vs extending the module's core. This is a perfectly acceptable preference and an approach that I am happy to take, however I feel because this module's core functionality is actually applying `.env` file values to `process.env`, the option to choose precedence over existing values is both worth considering for the module's core functionality and would definitely be a used option.

With that being said, my biggest reasoning for _why_ I feel this option is worth including is that **not all environments give the control to set environment variables** as [you state in #115](https://github.com/motdotla/dotenv/issues/115#issuecomment-168025406) for the current logic.

Take docker and docker compose for example; they [auto-create environment variables](https://docs.docker.com/compose/env/) based on container names. This is how docker exposes/links containers to one another, but it isn't always useful in the application's context – e.g. `name_PORT` returns a full URL and not a PORT as the name indicates.

So in this docker environment with a linked `db` container, the following application config fails because docker sets those environment variables prior to the application running regardless if you have the correct values defined in a dotenv file.
```
{
  host: process.env.DB_ADDR,
  port: process.env.DB_PORT,
  database: process.env.DB_NAME,
  user: process.env.DB_USER,
  pass: process.env.DB_PASS,
}
```

I know this could be tossed out as a docker issue (which I agree _is_ an issue with their naming convention), but doing so falls into a _not my problem_ mentality which rarely actually solves problems. Also, docker is not the only environment that auto-sets environment variables or restricts editing environment variable values.

By allowing the developer to choose if the `.env` file values take priority over machine set variables, it reduces external factors that can introduce bugs (other programs on the machine setting conflicting variables) and **allows for better application encapsulation with clean separation of code and config**.

---

Lastly, I'm not trying to force this option on the module (pun somewhat intended), but rather sincerely proposing a contribution that I feel aligns with the modules goals and solves a problem that other people (including myself) have encountered. If you have requested edits (such as a different option name) or simply think it does not fit in the module's core, just let me know.

Once again, I apologize for the length of my response and thank you for considering this pull request.

maxbeatty:
If I understood your position correctly, the primary use case for overriding environment variables is running your application inside a container such as docker. I would argue that if you're using docker, you shouldn't be using dotenv. Docker has an argument (`--env-file`) to load a file to populate environment variables. I agree Docker Compose is very opinionated with their naming conventions, but in my experience, database modules have been able to use the URL provided directly as opposed to defining an object with each part broken down. (If your module can't handle a full URI, I'd suggest parsing it with the built-in `url` module.)

It's a good discussion, but I'm still not sold. :smiley_cat:

galenandrew:
Yeah, we've used the env file option with docker compose. Docker only runs our local development but we still use this package since we have other hosted environments.

Based on your responses it seems as if you're looking for a perfect use case to justify merging this option. I'm not sure if there is ever going to be one that satisfies that requirement.

As an inverse to that approach, let's discuss reasons why this option is not a good fit? Does it conflict with the package goals or focus? Does it introduce issues with core functionality? Why should this _not_ be merged?

This is a good discussion, but I'm not sold it's a bad fit.

PR #124, opened 2016-01-26
Fix false-positive test due to case-sensitive process.env variables

galenandrew:
_Moved from #123_

This test was returning a false positive because the `readFileSyncStub` loads `test=val` (lowercase) and the test is checking for `process.env.TEST` (uppercase), thus never colliding with an existing env variable.


motdotla:
@galenandrew apologies for maybe being thick headed, but can you explain this one further. I don't understand.

PR #130, opened 2016-03-16
Add override option

eddiezane:
Been burned a few times by having a system environment variable and forgetting about or not knowing that dotenv would override it. Most people on my team set a global `TWILIO_AUTH_TOKEN` in their shell config which I'm sure is common with other developers.

This adds a backwards compatible feature to force environment variables in a `.env` file to override the system ones.

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,6 +102,16 @@ using this option.
 require('dotenv').config({encoding: 'base64'});
 ```
 
+#### Override
+
+Default: `false`
+
+You may specify that system environemnt variables should be overridden by those in a `.env` file.
+
+```js
+require('dotenv').config({override: true});
+```
+
 ## Parse
 
 The engine which parses the contents of your file containing environment
@@ -152,6 +162,8 @@ No. We **strongly** recommend against having a "main" `.env` file and an "enviro
 ### What happens to environment variables that were already set?
 
 We will never modify any environment variables that have already been set. In particular, if there is a variable in your `.env` file which collides with one that already exists in your environment, then that variable will be skipped. This behavior allows you to override all `.env` configurations with a machine-specific environment, although it is not recommended.
+
+**Note**: This behavior can be overriden by setting the `override` config option to `true`.
 
 ### Can I customize/write plugins for dotenv?
 
--- a/lib/main.js
+++ b/lib/main.js
@@ -12,6 +12,7 @@ module.exports = {
     var path = '.env'
     var encoding = 'utf8'
     var silent = false
+    var override = false
 
     if (options) {
       if (options.silent) {
@@ -23,6 +24,9 @@ module.exports = {
       if (options.encoding) {
         encoding = options.encoding
       }
+      if (options.override) {
+        override = options.override
+      }
     }
 
     try {
@@ -30,7 +34,7 @@ module.exports = {
       var parsedObj = this.parse(fs.readFileSync(path, { encoding: encoding }))
 
       Object.keys(parsedObj).forEach(function (key) {
-        process.env[key] = process.env[key] || parsedObj[key]
+        process.env[key] = override ? parsedObj[key] : process.env[key] || parsedObj[key]
       })
 
       return parsedObj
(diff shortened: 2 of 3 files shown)


motdotla:
Oh hello @eddiezane :) - we'll look at this soonish

devinrader:
I can see where overriding the environment vars might actually lead to bad things, so what if instead you at least printed out a warning that the same key was found in both places, that would help debug those situations where your wondering why you r code isn;t working and its just because you forgot you already have a variable defined.

motdotla:
Let's get others to weigh in here. My thoughts are:

+ dotenv should not override any previously set vars
+ I like the suggestion by @devinrader that it instead warn - to help debug this scenario

My thoughts here generally stem from the perspective that env is for setting application variables not user based system variables. Setting this twilio_auth_token is blurring that line. 

@eddiezane can you elaborate on why you set the env variable across your whole system - rather than application specific?

dkundel:
I agree that dotenv should be mainly for application based variables. But this actually makes Eddie's PR even more valid in my opinion. If you look at how Node tries to load modules they do the very same thing. Node will first look at your local `node_modules` folder and then afterwards into your global modules. 
The values that are in a project specific `.env` file are usually way more application specific and therefore, at least in my opinion, more valid for the project than global ones.

nimrod-becker:
My two cents on the matter, 
I agree that dotenv should not override previously set vars by default, however in some cases I do see the need to force it.

In my case, my processes are being run under a supervision mechanism (forever, supervisord etc.). In these cases, and maybe its a mis-behaviour of the supervision mechanism, the parent process is holding the env vars and the child processes inherits them. This means that in order to change values in .env and have the affecting the program, I need to take down the supervisor as well.

Forcing to load the vars in .env would easily solve this, however I might suggest a different approach. @motdotla , @devinrader  suggested issuing a warning with conflicting keys. How about supplying a callback function which will receive the conflicting keys and their values in both places? This way one can choose to simply warn, and another to selectively override.

eddiezane:
My thoughts are that the closer to the code you are the closer the truth should be. Per project configs should take precedence over global system configs. Since `.env` files are [discouraged](https://github.com/motdotla/dotenv#should-i-commit-my-env-file) from being committed to a project, if one exists it should be deliberate.

I took a at the [Ruby implementation](https://github.com/bkeepers/dotenv) and found a similar method of [overload](https://github.com/bkeepers/dotenv/blob/7b56b4e55261da8c5f7977b9bc1001a87ceb777f/lib/dotenv.rb#L30).

maxbeatty:
> If you look at how Node tries to load modules they do the very same thing. Node will first look at your local node_modules folder and then afterwards into your global modules.

This is only true "if the `NODE_PATH` environment variable is set to a colon-delimited list of absolute paths" according to [Node.js docs](https://nodejs.org/docs/latest/api/modules.html#modules_loading_from_the_global_folders). This exemplifies the danger I see in adding the capability to override environment variables. You could innocently pick a conflicting system variable name like `PATH` or `NODE_PATH` and your modules would stop loading. 

> I took a look at the Ruby implementation and found a similar method of overload.

I'm not concerned with feature parity of other tools by the same name in different languages.

--
This isn't something I'm interested in adding to the core of dotenv because it can lead developers to do unsafe modifications to the process. Additional debugging will need to be added for when people do bad things to themselves which means even more for us to maintain. This would be better off as an expansion module. I'm happy to add a link to your project for overriding environment variables to the [README](https://github.com/motdotla/dotenv#go-well-with-dotenv).

motdotla:
> This would be better off as an expansion module

I'm in agreement with @maxbeatty. Let's not put it in core.

@eddiezane would be rad if you'd roll this into its own library on top of dotenv. You can check dotenv-expand for pattern inspiration [1]

Then we can link to your repo in the README - similarly to how we link to dotenv-expand.

[1] https://github.com/motdotla/dotenv-expand

eddiezane:
Sounds good. Thanks for the collaboration all.

PR #134, opened 2016-04-06
Add cache option

kzima:
Allow signal events to update config on the fly.

### DIFF
--- a/test/main.js
+++ b/test/main.js
@@ -166,5 +166,15 @@ describe('dotenv', function () {
       parsed.INCLUDE_SPACE.should.eql('some spaced out string')
       done()
     })
+
+    it('should override basic environment if cache is set to false', function (done) {
+      var testPath = 'test/.env'
+      dotenv.config({path: testPath})
+      process.env.BASIC.should.eql('basic')
+      process.env.BASIC = 'test'
+      dotenv.config({path: testPath, chache: false})
+      process.env.BASIC.should.eql('test')
+      done()
+    })
   })
 })
--- a/README.md
+++ b/README.md
@@ -69,6 +69,16 @@ _Alias: `load`_
 
 ### Options
 
+#### Cache
+
+Default: `true`
+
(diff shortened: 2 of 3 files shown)


maxbeatty:
Can you expand the description? I'm not clear on what the use case is

kzima:
Imagine a situation where API key in .env file needs to be changed, but we don't want to restart the node server (this can be achieved with signals). Line 33 in lib/main.js prevents from doing this. So instead of completely removing first expression from `process.env[key] || parsedObj[key]` I just  added an extra cache option.

maxbeatty:
Sorry, I still don't understand how this works. How are you re-running `config` with a signal? Can you provide some example code here or in a gist?

kzima:
// listen to the process signal and restart the environment
process.on('SIGINT', () => {
  require('dotenv').config({ cache: false });
});

maxbeatty:
Can you add a test? This looks to function effectively the same as #130 😬 (`cache` set to `false` would allow overriding the current `process.env`)

kzima:
hey @maxbeatty I have added a test.

test/main.js
```diff
@@ -166,5 +166,15 @@ describe('dotenv', function () {
       parsed.INCLUDE_SPACE.should.eql('some spaced out string')
       done()
     })
+
+    it('should override basic environment if cache is set to false', function (done) {
+      var testPath = 'test/.env'
+      dotenv.config({path: testPath})
+      process.env.BASIC.should.eql('basic')
+      process.env.BASIC = 'test'
+      dotenv.config({path: testPath, chache: false})
```
jcblw: hmm is this test passing? If so it might be evergreen cause cache is spelled wrong.

maxbeatty: for whatever reason, travis did not build this commit

maxbeatty:
Thanks for the idea @kzima. I'm going to close this in favor of #130 since that's where more discussion about overriding is happening.

PR #137, opened 2016-04-12
added the override option in config function.

liangzan:
the override option allows the keys stated in .env to override the
keys that is present on the environment.

my case is i have several aws users with different permissions. i need the values in the dot env file to have higher precedence over those in the environment

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -12,6 +12,7 @@ module.exports = {
     var path = '.env'
     var encoding = 'utf8'
     var silent = false
+    var override = false
 
     if (options) {
       if (options.silent) {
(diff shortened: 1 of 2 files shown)


dkundel:
There is a conversation about this going on in #130. I think it would be cool if you could chip in your perspective there :)

maxbeatty:
> my case is i have several aws users with different permissions. i need the values in the dot env file to have higher precedence over those in the environment

I would suggest [loading credentials from disk](http://docs.aws.amazon.com/AWSJavaScriptSDK/guide/node-configuring.html)

closing in favor of #130

PR #138, opened 2016-04-15
Coveralls Intergration

jcblw:
@maxbeatty tried integrating this a while ago, we probably should've at the time. I was at js.la and @bcoe recommended it :tada:


jcblw:
@motdotla

jcblw:
@motdotla you can turn these off in the settings of coveralls.io, I does not seem I have access to do it.

bcoe:
\o/ hooray the coverage movement marches forward.

Issue #140, opened 2016-05-02
dotenv.config() truncates leasing '='


liorshalev01:
For example:
`PASSWORD="=211cb1cdb045df37I"`
The leading '=' is required for the correct password input.

jcblw:
interesting, have not seen this before. I was able to reproduce with `dotenv 2.0.0` and `node 6.0.0`

```shell
# dotenv test .env
PASSWORD="=211cb1cdb045df37I"
```

and the command

```shell
node -r dotenv/config -e "console.log(process.env.PASSWORD)"
```

jcblw:
I wrote some test in #141 I was only able to replicate the bug when using the `-r` preloading option. @liorshalev01 are you using the preload or at the top of a file?

jcblw:
After some more testing I noticed that 

```
TEST='=FOO' node -e 'console.log(process.env.TEST)'
```
will output `FOO` as well. So it looks like there is some processing that is happing via node that is removing the leading `=`. I would strongly recommend not relying on this to work in the time being so you do not run into issues going from dev to production and trying to set those just in the machine's environment. If you know this equal sign will always prefix the key I would add it to the key post the env being loaded.

jcblw:
filed this ticket for this issue https://github.com/nodejs/node/issues/6627

jcblw:
ok so the feedback nodejs/node#6627 is that this is a side-effect from the shell your using @liorshalev01. I was only able to replicate the bug while using `ZSH`. Can you confirm that the leading `=` is retained while using `bash`.

eg.
```shell
bash -l
node start # or how ever you start your app
```

liorshalev01:
Hi,
I had some tests and I can confirm that it might be the shell side effect.
I was able to reproduce it only OSX.
On Ubuntu/Linux it works (e.g., no leading '=').

Thanks,
Lior

jcblw:
Awesome we are not going to be able to do much about this in the dotenv thanks for the feedback.

PR #141, opened 2016-05-07
leading equal bug

jcblw:
It seems as tho adding this in did not fail. This is to help address #140


jcblw:
Interestingly enough, this test will fail in `ZSH` but works in `BASH`. Not including due to that fact.

PR #144, opened 2016-06-13
Try to locate .env in parents

zatziky:
If .env is not found in current folder and no `options.path` is specified then try to locate parents.

Would you accept this feature? I can make optional. I will refactor the code according to your guidelines and cover the lines with tests if we will agree on adding this feature. Let me know.

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -5,7 +5,7 @@ var fs = require('fs')
 module.exports = {
   /*
    * Main entry point into dotenv. Allows configuration before loading .env
-   * @param {Object} options - valid options: path ('.env'), encoding ('utf8')
+   * @param {Object} options - valid options: path ('.env'), encoding ('utf8') findInParents (false)
    * @returns {Boolean}
   */
   config: function (options) {
@@ -23,6 +23,10 @@ module.exports = {
       if (options.encoding) {
         encoding = options.encoding
       }
+    }
+
+    if (!options || options && !options.path){
+      path = _findFileInParents(path, encoding, silent);
     }
 
     try {
@@ -77,6 +81,24 @@ module.exports = {
     return obj
   }
 
+}
+
+function _findFileInParents(path, encoding, silent) {
+  for (var i = 0; i < 100; i++) {
+    try{
+      fs.readFileSync(path, { encoding: encoding })
+      return path
+    }
+    catch (e){
+      path = '../' + path
+    }
+  }
+
+  if (!silent) {
+    console.error("No .env found.");
+  }
+
+  return false;
 }
 
 module.exports.load = module.exports.config


maxbeatty:
Is the motivation behind this anything more than being lazy and/or misinformed (i.e. I don't know where the .env file is)?

zatziky:
Yes, we run lot's of small scripts independently. It's data mining scripts. We have almost 20 now and there will be more of them. Every script needs the same config from .env but it's clumsy to set `options.path` in different folder depths and it's not refactoring safe. 

I bet more teams would appreciate this feature. Of course, I can always write a method, e. g.:

`require('dotenv').load({path: resolveEnvPath()});` 

I'd say it could fall in responsibilities of the module.

maxbeatty:
Sounds like you may want a more robust configuration tool like [consul](https://www.consul.io/). dotenv really isn't intended to be used across multiple projects, more like one `.env` per project. 

> Every script needs the same config from .env but it's clumsy to set options.path in different folder depths and it's not refactoring safe

Symlinks are another workaround or even using `process.env`

```js
require('dotenv').load({ path: process.env.CONFIG_PATH })
```

```
CONFIG_PATH=~/.env node index.js
```

I'll leave some code comments outlining why this may be harder than first pass

lib/main.js
```diff
@@ -79,4 +83,22 @@ module.exports = {
 
 }
 
+function _findFileInParents(path, encoding, silent) {
+  for (var i = 0; i < 100; i++) {
+    try{
+      fs.readFileSync(path, { encoding: encoding })
+      return path
+    }
+    catch (e){
```
maxbeatty: need to validate the exception is because file does not exist. would be better to not read file twice. check out `fs.exists`

lib/main.js
```diff
@@ -79,4 +83,22 @@ module.exports = {
 
 }
 
+function _findFileInParents(path, encoding, silent) {
+  for (var i = 0; i < 100; i++) {
```
maxbeatty: ideally would stop at root of file system

lib/main.js
```diff
@@ -79,4 +83,22 @@ module.exports = {
...
+      fs.readFileSync(path, { encoding: encoding })
+      return path
+    }
+    catch (e){
+      path = '../' + path
+    }
+  }
+
+  if (!silent) {
+    console.error("No .env found.");
```
maxbeatty: can omit since already done in `config`

lib/main.js
```diff
@@ -79,4 +83,22 @@ module.exports = {
...
+  }
+
+  return false;
```
maxbeatty: why `false`? how will it be used?

lib/main.js
```diff
@@ -79,4 +83,22 @@ module.exports = {
...
 }
 
+function _findFileInParents(path, encoding, silent) {
+  for (var i = 0; i < 100; i++) {
+    try{
+      fs.readFileSync(path, { encoding: encoding })
+      return path
+    }
+    catch (e){
+      path = '../' + path
```
maxbeatty: biggest issue: paths need to be platform agnostic. you'll most likely want to use `path.resolve` but you'll need a starting point like `process.cwd()` which runs the risk of breaking people's current usage. needs lots of validation it works as expected

zatziky:
Consul is too heavy but thank you for an alternative.

`process.env.CONFIG_PATH` would work, however, it is exactly what we'd like to avoid. That each team member would have to set env. vars.

lib/main.js
zatziky: What could be the risk with `process.cwd()`? One thing I am not very sure would be how to stop at the root of an OS. 

Thanks for the comments. As I said if we agree on adding this feature I will improve the code. Let me know Max.

maxbeatty: I would like to see more demand from the community and a complete PR including tests and documentation before seriously considering this addition

j-klassen:
I have a use case in the same ball park with [ava](https://github.com/avajs/ava). Currently they are changing `process.cwd` per test script (sounds like this will change), causing `dotenv` to fail to load the `.env` file. To get around this I used [find-up](https://github.com/sindresorhus/find-up). This way you could make your own script/module that simply is required before `dotenv/config`:

```javascript
// dotenv-argv.js

const findUp = require('find-up');
process.argv.push(`dotenv_config_path=${findUp.sync('.env')}`);
```

```javascript
// package.json
...
"scripts": {
  "test": "ava -r ./lib/dotenv-argv -r dotenv/config"
}
...
```
I realize this isn't exactly the same situation, but it's an issue I've encountered using `dotenv` with other modules.

maxbeatty:
@j-klassen thanks for calling out `find-up`. It should solve @zatziky's need quite well

```js
const findUp = require('find-up')
dotenv.config({
  path: findUp.sync('.env')
})
```

zatziky:
@j-klassen and @maxbeatty, guys, thank you! It is just what I needed.

jamesdixon:
@j-klassen thanks for this -- was running into the same issue with Ava 👍

PR #147, opened 2016-08-04
added assignToProcessEnv to options

sean-shirazi:


maxbeatty:
You can already do this by [calling `parse` directly](https://github.com/motdotla/dotenv/blob/1e73f004c6783592a1a014d7077f5112ea7c5f72/README.md#parse)

sean-shirazi:
Here is my use case. I found the first code cleaner than the second:
``` js
const obj = require('dotenv').config({
    assignToProcessEnv: false,
}) || {};
```

``` js
const fs = require('fs');
try {
    var obj = require('dotenv').parse(fs.readFileSync('.env', {
        encoding: 'utf8'
    }))
} catch (err) {
    console.error(err)
    var obj = {}
}
```

maxbeatty:
We expose the `parse` method so people can use it for things other than populating `process.env` (the main purpose of this module). You're able to accomplish your goal without code changes on our end which is more important to me as a maintainer than code cleanliness on your end. Thanks for understanding 😸

sean-shirazi:
alright thanks :-)

PR #148, opened 2016-08-07
Update README for consistent semicolon usage

djds23:
The current README is perfectly fine, however the lack of consistency was distracting.

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,7 +113,7 @@ an Object with the parsed keys and values.
 var dotenv  = require('dotenv');
 var buf    = new Buffer('BASIC=basic');
 var config  = dotenv.parse(buf); // will return an object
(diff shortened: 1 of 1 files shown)


maxbeatty:
We use [Standard Style](http://standardjs.com) which does not use semicolons

jcblw:
@djds23 if you like, make a PR removing the semicolons, and that I more likely to be accepted.

maxbeatty:
What would be super awesome is to lint the JS code blocks in Markdown documents to keep everything consistent

jcblw:
Yeah i have not seen that before, I sure its out there tho.

maxbeatty:
😉 https://github.com/hapijs/lab/pull/513

Issue #135, opened 2016-04-07
Error: ENOENT: no such file or directory, open '.env'


gragra100:
Hi,
I am working on a Ubuntu 14 Virtual Machine. If i run my `node server.js` inside the project folder that contains my .env file it's all ok! BUT if i run for example `node /home/myapp/server.js` it return this error:

> { [Error: ENOENT: no such file or directory, open '.env'] errno: -2, code: 'ENOENT', syscall: 'open', path: '.env' }

and then my server starts on a random port.
Can anyone help me? I need to run the server on startup so i need to add a .conf file on /etc/init with the `node /home/myapp/server.js` command.

jcblw:
I think we are just logging out this error, your server should be running fine. If you want to get rid of the log you can pass the option of `silent: true`.

```javascript
require('dotenv').config({silent: true})
```

 Im assuming your setting your port via `.env` all you would need to do on your server is set that process variable in you upstart script or something like that 

```
export PORT=5000
```

maxbeatty:
By default, the `.env` file is loaded from [`process.cwd()`](https://nodejs.org/docs/latest/api/process.html#process_process_cwd). When you run `node server.js`, the current working directory is `/home/myapp/` so your file is found. When your upstart script runs or you run `node /home/myapp/server.js`, the current working directory is whatever `pwd` returns so your file isn't found.

Here are two ways to work around this:

1. configure [path](https://github.com/motdotla/dotenv#path) to `/home/myapp/.env`
2. use something like [`pushd`](https://en.wikipedia.org/wiki/Pushd_and_popd) to set your directory stack in your init script (e.g. `pushd /home/myapp >& /dev/null && node server.js`)

caggie:
`require('dotenv').config({path: __dirname + '/.env'});`

Use this and you can run `node /home/myapp/server.js`

samson212:
`__dirname` won't work if your app isn't run directly with node, i.e. you're using `systemctl`. I had to use `process.env.HOME` instead.

battalian:
I am getting this error when i tried to install cordova & ionic "sudo npm install -g cordova ionic", what should i do??

![image](https://user-images.githubusercontent.com/18572376/27468284-9edc5236-5806-11e7-881b-0ef35b0f6606.jpg)

jesmarquez:
I have the file env in directory root. I get this message:

Error: ENOENT: no such file or directory, open '.env'

The value process.cwd() is correct.

[...]

PR #154, opened 2016-10-18
suppress error messages by default

maxbeatty:
logging errors by default is scaring more people than it's helping. see #135 and [this travis output](https://travis-ci.org/jsperf/jsperf.com/builds/168675049#L1189) as examples

this should result in a major release


maxbeatty:
@motdotla @jcblw do you agree logging the errors is doing more harm than good?

jcblw:
I agree that this should be turned off by default.

heitortsergent:
@maxbeatty are there any reasons to drop support for older versions of Node?

maxbeatty:
Nothing significant is changing about the implementation that would prevent it from working, but those older versions are no longer receiving updates like security patches so I think it's better for the community at large to go forward. If you're using a sane semver range, you won't notice this.

heitortsergent:
Based on the table [here](https://github.com/nodejs/LTS#lts-schedule), I would suggest leaving 0.12 support and opening a separate issue/PR for dropping support for it in Dec/Jan. That seems to be the timeline for ending critical patches to it.

jcblw:
@maxbeatty  @motdotla  we looking to do a major bump with this? If so and you guys think #158 is cool we should probably get that in the bump as well.

maxbeatty:
yes, should be a major version release because of the breaking change to the silent option

PR #155, opened 2016-10-22
Close #115. Add example code for overriding process.env to FAQ

### LINKED ISSUE #115: add 'replace' option to overwrite existing process.env keys

Glavin001:
See https://github.com/motdotla/dotenv/issues/115#issuecomment-168210754


jcblw:
awesome thanks for this 👍

Issue #156, opened 2016-10-29
Should allow comment at end of lines


mettjus:
I think it would be useful to be able to have comments at the end of lines. E.g.:

```
MY_VAR=some-value   # this should be treated as a comment
```

jcblw:
I like this idea, I would be open to merging a PR for this.

mettjus:
I played a little with it and was wondering if you had already taken into account the opportunity of using a parser generator with an expression grammar, something like peg.js. I experimented with it and already implemented the comment-at-end-of-line feature using peg.js. If you want to take a look you can find it [here](https://github.com/mettjus/dotenv/tree/end-of-line-comment).
The object resulting from parsing the test .env file is the same.
Let me know what you think, if this is an acceptable approach or if you'd rather stick to a manually defined parser.

maxbeatty:
Why are comments needed inline? Please expand on its usefulness.

sneakertack:
They probably don't qualify as needed, maybe they're more like a nice-to-have: in certain cases some developers might prefer being able to write a short comment on the same line, instead of having to add a new line above (unfortunately I don't have stats, I can only anecdotally say I've had that want).

My specific context was during the preparation of a `.env.sample` file with 8 or so KV pairs (that teammates could `cp` and then make minor edits to). They're sample values, so I can mostly get by by describing the expected value using the value itself (e.g. `POSTGRES_URL=postgres://user:password@host:port/5432`). However for 1 or 2 pairs it would have been nice to provide a bit more info (e.g. ` # append ?ssl=true when connecting to heroku`), yet dedicating a separate line for the comment would have felt excessive.

Shell script allows #-comments midway through the line, and it's plausible that this module might consume environment variable definition files originally written for Shell, so for me inline comments being unsupported rather than supported was the part that felt surprising.

maxbeatty:
dotenv doesn't need to support inline comments for a sample file to get its point across. If there's mass confusion, use the comments that are already supported. `.env` should be thought of as a simple ini file and nothing more

sneakertack:
Okay. In case anyone else is specifically looking for this feature in future, you can use this fork here: [sneakertack/dotenv#altparser](https://github.com/sneakertack/dotenv/tree/altparser).

Additional context can also be found at #158, be sure you understand the differences before installing.

PR #158, opened 2016-11-25
End-of-line comments, quoted whitespace, and other combination edge cases

sneakertack:
This PR implements the following changes:

1) Allow comments at the end of lines, like mentioned in #156
2) [Maybe breaking] no longer trim leading/trailing whitespace if it's surrounded by quotes (the assumption being if the user is including whitespace in quotes, it's probably what they want)

The main motivation is (1), as I too would like to be able to comment at the end of lines; but I suspect that the addition of a PEG.js parser might be a bit too heavy-handed (adds >600 new SLOC) when this library should be lightweight/zero-dependency.

It has 4 commits:

- The 1st commit adds tests for EOL-comments (initially failing), as well as some test cases for surrounding whitespace (quoted test fails - trimming is too aggressive)
- The 2nd commit tries to do the simplest possible change to the `parse` function (modify the regex + remove `trim()`). It almost works, however, 1 test doesn't pass (it fails in the tricky case of combining quotes and the comment hash character). This hopefully justifies why I had to resort to making a bigger change next.
- The 3rd commit changes the logic of the `parse` function to better handle quotes + hashes. The prior failing test now passes.
- The 4th commit adds even more edge-case tests (they pass), and updates the readme to reflect the new feature.

All tests are passing in my local environment. I've also run it against the linter. Discussion and requests for amendment are welcome.

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,7 +125,7 @@ The parsing engine currently supports the following rules:
 - empty lines are skipped
 - empty values become empty strings (`EMPTY=` becomes `{EMPTY: ''}`)
 - single and double quoted values are escaped (`SINGLE_QUOTE='quoted'` becomes `{SINGLE_QUOTE: "quoted"}`)
-- leading/trailing whitespace gets trimmed, unless within quotes
+- leading/trailing whitespace gets trimmed
 - new lines are expanded if in double quotes (`MULTILINE="new\nline"` becomes
 
 ```
--- a/README.md
+++ b/README.md
@@ -123,9 +123,9 @@ The parsing engine currently supports the following rules:
 
 - `BASIC=basic` becomes `{BASIC: 'basic'}`
 - empty lines are skipped
-- lines beginning with `#` are treated as comments
 - empty values become empty strings (`EMPTY=` becomes `{EMPTY: ''}`)
 - single and double quoted values are escaped (`SINGLE_QUOTE='quoted'` becomes `{SINGLE_QUOTE: "quoted"}`)
+- leading/trailing whitespace gets trimmed, unless within quotes
 - new lines are expanded if in double quotes (`MULTILINE="new\nline"` becomes
 
 ```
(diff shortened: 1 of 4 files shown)


maxbeatty:
This would be better as two separate pull requests– one to address inline comments and another to address trimming whitespace in quotes.

In regards to inline comments, I think it adds more complexity than usefulness. No one has presented any use cases. Why do you need a comment about a key value pair that can't go on the line above?

In regards to trimming whitespace, is this actually a problem? If lots of people are having their values incorrectly parsed, let's fix it, but if no one has noticed this for years, let's leave it alone.

sneakertack:
Firstly, apologies for conflating the 2 issues into the same PR. The whitespace change was a secondary issue encountered while trying to implement EOL comments, and I bundled it because their code lines weren't fully independent. In an event where you'd like to go ahead with 1 change but not the other, I'll be happy to issue a new PR implementing only the approved change.

I've replied about inline comments in its original issue, so maybe we can continue with whitespace here.

On my end, my reasoning for proposing to change the whitespace treatment was because I was assuming that 1) more specification, rather than less, about how the module's features interact (e.g. quotes vs. whitespace handling) is generally better, and 2) in that case, the current specification for how whitespace gets trimmed even when enclosed within quotes does seem a bit odd.

1 case I can come up with (post-hoc) for the change is the event whereby whitespace is meant to be the actual value being configured, where given the current implementation they would end up get trimmed to an empty string:

```shell
INDENT_PADDING="  " # 2 spaces
DELIMITER="	" # tab character
```

But yes, I admit that this situation may be unlikely, to the point where no one using the module in its history has encountered it yet.

jcblw:
@sneakertack I would have to agree with @maxbeatty on the "white space in quotes" issue. I don't think we should bundle that into this PR. I think it added some extra complexity that is not needed to achieve the EOL comments parsing. 

I would say try not to open a new PR, try to revert those lines. It is always good to keep a accurate history on a PR.

sneakertack:
Okay, I've reverted the quoted-whitespace change and updated the tests/readme accordingly. It doesn't necessarily decrease the code line changes in the `parse` fuction though (at least not for the implementation I went with) - while trying to develop the feature I struggled more with quotes vs. hash-character, rather than quotes vs. whitespace, so @maxbeatty is right in that the EOL-comment parsing is the bit that adds more complexity.

maxbeatty:
sorry, this really breaks people's expectations of how dotenv should work out of the box (like any password with a # is going to get truncated if it's not quoted). low value feature + additional code complexity + breaking expectations ≠ good

PR #163, opened 2016-12-05
Changed export statement to support ES6 import

jorrit:
This change allows importing config like this:
`import { config } from 'dotenv'`


jorrit:
Thanks!

PR #164, opened 2016-12-17
Added info about silent mode to README.md

Hakier:


maxbeatty:
thanks but the [verbose option](https://github.com/motdotla/dotenv/blob/8117fc1234805323b1d8c05a2ea79d1fc3675255/README.md#verbose) is already documented. see #154 for the switch from `silent`

PR #165, opened 2016-12-20
Add support for throwing if failed

IOAyman:
If `verbose: 'throw'` is passed in options, an `Error` is thrown instead of `console.error`ing to the stdout

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,15 @@ All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
+### Changed
+
+- Return Object with parsed content or error instead of false ([#165](https://github.com/motdotla/dotenv/pull/165)).
+
+
+### Removed
+
+- `verbose` option removed in favor of returning result.
+
 
 ## [3.0.0]
 ### Added
--- a/README.md
+++ b/README.md
@@ -65,21 +65,12 @@ $ node -r dotenv/config your_script.js dotenv_config_path=/custom/path/to/your/e
 
 _Alias: `load`_
 
-`config` will read your .env file, parse the contents, and assign it to
-[`process.env`](https://nodejs.org/docs/latest/api/process.html#process_process_env). You can additionally, pass options to
-`config`.
+`config` will read your .env file, parse the contents, assign it to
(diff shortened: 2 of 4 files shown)


maxbeatty:
I think a better alternative would be to return the error instead of false so consumers can handle the error as they please (throw, report, ignore, etc.)

IOAyman:
Yeah, it's better than returning `false`.
How about supporting both approaches? Returning the `Error` by default, but give the ability to throw if explicitly required via the options.

maxbeatty:
Please correct me if I'm wrong. The motivation to add an option to throw the error is to control how the error is handled.

Instead of adding options, perhaps it's better to remove one and change what's returned from `config`. It keeps the footprint and complexity of this module small while allowing the most flexibility for consumers. Taking a page from [Hapi's Joi module](https://github.com/hapijs/joi/blob/045b99c98d28ec72edf7749d3adb4c99e18c610a/API.md#errors), we could return an object with an `error` key that is either `null` or whatever error occurred and a `parsed` key with the result of parsing the .env file.

```js
import { config } from 'dotenv'
import variableExpansion from 'dotenv-expand'

env = config()

if (env.error) {
  // handle error as you like
  throw env.error
}

variableExpansion(env.parsed)
```

This is a low-impact path forward for anyone upgrading major versions. It allows people to continue ignoring errors that are most likely due to a missing file in environments like CI.

- [ ] remove `verbose` option
- [ ] change return value of `config`
- [ ] update README
- [ ] update CHANGELOG

jcblw:
This looks good my only concern is with the the return object change and external plugins that do not have `dotenv` as a dep, eg. [dotenv-expand](https://github.com/motdotla/dotenv-expand), which has no control if `v2` or `v3` of `dotenv` is used.  

I do not think there is many of these plugins out there, but I do think it would be a good idea, if this is merged, that we make either PRs or issues to update to support both versions. So that way on release we are not breaking those plugins functionality.

maxbeatty:
other modules should specify [peer dependencies](https://docs.npmjs.com/files/package.json#peerdependencies) to indicate which versions of dotenv they are compatible with

motdotla:
> other modules should specify peer dependencies to indicate which versions of dotenv they are compatible with

I agree.

:+1: for this PR

PR #166, opened 2016-12-24
Include comment after variable definition (partial comments).

jourdanrodrigues:


lib/main.js
```diff
@@ -13,7 +13,7 @@ function parse (src) {
   // convert Buffers before splitting into lines and processing
   src.toString().split('\n').forEach(function (line) {
     // matching "KEY' and 'VAL' in 'KEY=VAL'
-    var keyValueArr = line.match(/^\s*([\w\.\-]+)\s*=\s*(.*)?\s*$/)
+    var keyValueArr = line.replace(/( +)?#(.+)?/, '').match(/^\s*([\w\.\-]+)\s*=\s*(.*)?\s*$/)
```
jourdanrodrigues: I did not use `\s` instead of the literal whitespace because the `\s` covers other "space" characters that should not be considered (very specific cases).

maxbeatty:
Sorry, I'm still not convinced of the usefulness of this especially when weighing the additional complexity and potential for unintended behavior.

jourdanrodrigues:
Hi, @maxbeatty. Saw later that you're against this idea and I understand your concern, but I think the complexity of this feature lies on the documentation to explicitly define it and tests to contain this unintended behaviors.

In this pull request, all is needed is to put a whitespace character before the hash (perhaps a specified number of whitespaces, like 3 or 4?).

But as I told before, knowing your thoughts about this feature, I thank for your attention. 😄

maxbeatty:
The `.env` format loosely follows [INI file](https://en.wikipedia.org/wiki/INI_file) which doesn't come with features like types, variable interpretation, or inline comments (previous feature requests). The primary value of dotenv is populating `process.env` with values from a simple `.env`. Defining and maintaining our own file format is well beyond the scope of the project.

If you really want those extra features for your config file, I'd suggest using a more robust file format for your configuration. For example, you could build "dotyml" in a few lines:

```js
const ryml = require("read-yaml")

module.exports = function (ymlFile) {
  try {
    const config = ryml.sync(ymlFile);
    Object.keys(config).forEach(function (key) {
        process.env[key] = process.env[key] || config[key]
    })
    return { parsed: config }
  } catch (e) {
    return { error: e }
  }
}
```

```js
require('dotyml')('path/to/env.yml')
console.log(process.env.your_var)
```

See it in action: https://runkit.com/586163c32869ba001432cdb8/586163c32869ba001432cdb9

PR #170, opened 2017-01-05
Add option 'proc': when set to false it will not modify process.env.

ckiickA:
Sometimes you want to only read in environment variables without modifying process.env.  This is useful when reading a .env file prior to spawning a new process, for example.  The 'proc' option defaults to true.  If set to false, it will not modify process.env, just return the parsed object.

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -45,6 +45,7 @@ function parse (src) {
 function config (options) {
   var path = '.env'
   var encoding = 'utf8'
+  var proc = true
 
   if (options) {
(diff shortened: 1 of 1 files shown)


maxbeatty:
the [parse](https://github.com/motdotla/dotenv/blob/v4.0.0/README.md#parse) method is exposed to do just that

```js
const fs = require('fs')
const parsedObj = dotenv.parse(fs.readFileSync('.env', { encoding: 'utf8' }))
```

PR #177, opened 2017-01-31
Parse lines beginning with "export"

atomaka:
We commonly attempt to reuse our `.env` files by including `export ` at the beginning of each line. This allows our `.env` files to also be sourced and loaded into our environment if necessary.  Similar dotenv packages seem to allow this functionality (ie. https://github.com/bkeepers/dotenv#usage).

This pull request maintains all existing functionality.  It adds one additional change.  A line formatted as

```
export EXPORTED=true
```

will be parsed with key `EXPORTED` and value `true`.


atomaka:
Looks like this functionality was proposed and declined in #118.

PR #179, opened 2017-02-20
update readme for react apps

mrdougwright:
This tripped me up more than once, so I thought I'd suggest updating the readme. Thanks for your time :)


maxbeatty:
this is an implementation detail of react and I worry that by this example you're sending sensitive information like database connection details to the client 🙁 

furthermore, I don't think dotenv should be encouraged for client use. other build tools like webpack includes [plugins that do the same thing](https://github.com/webpack/docs/wiki/list-of-plugins#defineplugin)

PR #181, opened 2017-03-05
Use dotenv as a cli command (like cross-env)

mickaelvalmier:
I need this feature for web bundler usage (eg: broswerify + envify) for inject environment variables forwarded by .env file.

Thanks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,6 +124,32 @@ line'}
 ```
 - inner quotes are maintained (think JSON) (`JSON={"foo": "bar"}` becomes `{JSON:"{\"foo\": \"bar\"}"`)
 
+## Command line interface
+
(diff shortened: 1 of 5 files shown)


maxbeatty:
I think this would be better as its own module. I was going to suggest naming it "[dotenv-cli](https://github.com/caske33/dotenv-cli)" but looks like someone is already wrapping this module to make a CLI. Happy to link to one or both in the README

mickaelvalmier:
My bad, i've not checked if a similar module exists.

But in my point of vue, import a cli in the same module is a better way than a separates repositories.

I usually use dotenv for client side and server side development, and is a convenient way to use the same module (as an universal development).

Moreover, it is easier to manage and follow the evolution of this feature like that.

I've fixed tests coverage, It's up to you if this addition it is worth it

See ya !

PR #188, opened 2017-04-12
Add json file support

assafshp:
Allow config method to support json file path to include environment variables

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -45,10 +45,14 @@ function parse (src) {
 function config (options) {
   var path = '.env'
   var encoding = 'utf8'
+  var parseFunc = parse
 
   if (options) {
     if (options.path) {
       path = options.path
+      if (options.path.toLowerCase().endsWith('json')) {
+        parseFunc = JSON.parse
+      }
     }
     if (options.encoding) {
       encoding = options.encoding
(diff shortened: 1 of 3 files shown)


maxbeatty:
Thanks but I think we'd like to stick to our single INI file format. It's easy to emulate dotenv with a few lines of code if you would like to use JSON:

```node
const myEnvConfig = require('.env.json')
for (const key in myEnvConfig) {
  process.env[key] = process.env[key] || myEnvConfig[key]
}
```

trainerbill:
@maxbeatty  not sure if this is still up for discussion or not, but I would like this functionality as well.  I am using Yeoman and it would be nice to be able to use readJSON and change the dotenv configuration in subgenerators.  I also use node -r dotenv/config when starting the application.

Your code does work but then I have to carry around an envConfig.js file in all my packages/projects.  With the widespread use of dotenv it would be nice to support this.  Otherwise I can just publish a module to do the exact same thing with JSON.

Let me know.  Thanks!

maxbeatty:
I'm still in the camp of publishing lots of small, single-purpose modules as opposed to offering more and more configuration of an existing module. I published [dotenv-json](https://www.npmjs.com/package/dotenv-json) to get most of the way there. Feel free to contribute or fork.

PR #189, opened 2017-04-18
parse `export KEY=VAL` so `.env` can be shared with shell scripts

iamdustan:
We have a `.env` file that uses the following syntax:

```sh
export LOCAL_KEY="local-value"
```

This adds support for the parse function to handle preceding `export ` keywords in each line.


maxbeatty:
this has been proposed before and I think we're happy with the current ini format

metaskills:
Thanks, but I'd like to say I'd be happier sharing with other scripts.

PR #214, opened 2017-07-14
Allow optional export keyword

### LINKED ISSUE #213: Allow for optional `export` at the beginning of each line
Sometimes it is useful to export the contents of the `.env` file into the current terminal session so that individual scripts can be tested in isolation. The only way it is possible right now is to use a rather convoluted script to read the contents and export them:

```
$ export $(cat .env | grep -v ^# | xargs)
```
It would be nice to allow for an optional `export` word at the beginning of each line:

```
export FOO=bar
```

This would allow users to `source` their environment vars into the current bash/terminal session:

```
source .env
```

DesignByOnyx:
Closes #213 
This was easy enough to go ahead and implement. I wont be offended if it's not accepted.


maxbeatty:
Per [comment](https://github.com/motdotla/dotenv/issues/213#issuecomment-315440784) in #213, I still think a module that extends core dotenv is the way to go for this

metaskills:
Using this now. https://github.com/sneakertack/nvar

PR #223, opened 2017-09-03
cut export and export arguments

slavaGanzin:
Now you can use dotenv with .env files containing export statements.
You don't need this in bash, but in fish statements isn't imported without set.


maxbeatty:
Sorry, we've discussed this in the past and this is not a feature we're looking to support.

PR #231, opened 2017-10-13
add overwrite option

dschaller:


maxbeatty:
Thanks but this has been suggest and discussed in the past. Overwriting environment variables still can have unintended side effects.

dschaller:
@maxbeatty I completely understand that it can have unintended side effects but I'm curious why not give people the option to do this if they understand these potential side effects?

This package will continue to work the same by default. This PR enhances this package by adding flexibility for those who need and understand what they are overwriting.

PR #236, opened 2017-11-01
Feature: ES2015

marcobiedermann:
Translate syntax to ES2015 because it is supported in Node 4.5+ and makes the code cleaner, faster and more maintainable

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -1,6 +1,4 @@
-'use strict'
-
-var fs = require('fs')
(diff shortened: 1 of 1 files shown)


maxbeatty:
Thanks. I included most of these changes in #237

marcobiedermann:
@maxbeatty 
Oh okay. I did not see that. Would be better to create an issue only for one particular feature to make it easier to find and available to other developers.

👍

PR #237, opened 2017-11-02
use explicit full path to form more informative error messages

maxbeatty:
I've noticed quite a few issues and comments where it's hard to debug when a `.env` file isn't found. By using `process.cwd` explicitly, the full path will be included in the error message to aid in debugging. Hopefully this is an improvement from having to work backwards from a custom path to figure out what the original error message meant.

- [x] start testing Node v9 and drop v7 since it's no longer maintained
- [x] use `process.cwd` to declare a full path to read from
- [x] remove references to iojs
- [x] introduce some es6 syntax ref #236


maxbeatty:
I might suggest this be released as a MAJOR version since there's a slim possibility that some assumptions could break by changing the default path that is being read.

maxbeatty:
@motdotla will you configure AppVeyor for the repository to enable Windows testing? I think you need to do it as the repo owner. This will unblock #241 and a new release (https://github.com/motdotla/dotenv/projects/1)

maxbeatty:
I am going to merge this and release v5. #260 has been created to track adding AppVeyor

PR #247, opened 2017-11-28
Return boolean values instead of "true"/"false" strings.

becdot:
When boolean values are set in an .env file (e.g. `IS_PRODUCTION=false`), it's expected that these will be evalulated as booleans, instead of the string `"false"` (which is actually truthy).  This causes particular problems when using webpack and webpack.DefinePlugin, since providing `JSON.stringify(process.env.IS_PRODUCTION)` will evaluate as truthy, which is unexpected.

Instead, change `"true"` to `true` and `"false"` to `false` when parsing the .env file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,7 +67,7 @@ _Alias: `load`_
 
 `config` will read your .env file, parse the contents, assign it to
 [`process.env`](https://nodejs.org/docs/latest/api/process.html#process_process_env),
-and return an Object with a `parsed` key containing the loaded content or an `error` key if it failed.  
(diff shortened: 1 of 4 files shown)


maxbeatty:
Thanks but we keep them as strings to align with how `process.env` behaves

> Assigning a property on `process.env` will implicitly convert the value to a string.
> – [Node.js](https://nodejs.org/api/process.html#process_process_env)

[envalid](https://github.com/af/envalid#validator-types) will do type conversion for you or you could write your own module focused on boolean behavior.

```js
dotenv.config()
convertBooleans(process.env)
```

becdot:
That makes sense -- thanks for the explanation!

PR #259, opened 2018-01-28
Added functionality for setting multiple paths.

mkovel:


maxbeatty:
This is in conflict with our guidance that [you should only have a single .env file](https://github.com/motdotla/dotenv/blob/6b07ace96fc39c865d4a2a60809eb0b97474e92b/README.md#should-i-have-multiple-env-files). If you must use multiple paths, you can handle this yourself 

```js

['.env', '/custom/path/to/your/env/vars'].forEach(path => dotenv.config({ path }))
```

mkovel:
@maxbeatty Hi, let me describe the main idea. 
My change still has the one .env file, but it can place in different places. 
I stop processing after than I found the first file.
It is not the same that you provided above.

PR #262, opened 2018-01-30
Refactor code to ES5

kevinhermawan:
I try to refactor ES6 code into ES5 code because I got an error when trying to build react app with create-react-app

### DIFF
--- a/config.js
+++ b/config.js
@@ -1,11 +1,13 @@
-(function () {
-  var options = {}
-  process.argv.forEach(function (val, idx, arr) {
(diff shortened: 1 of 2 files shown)


maxbeatty:
You should not be including dotenv in client bundles as it relies on the file system and populates `process.env` which is a Node.js construct. Instead, it may be useful _during_ the build process as [create-react-app is doing here](https://github.com/facebook/create-react-app/blob/25184c4e91ebabd16fe1cde3d8630830e4a36a01/packages/react-scripts/config/env.js#L44-L46)

PR #263, opened 2018-01-30
Transpile with babel. Closes #261

### LINKED ISSUE #261: Publish dotenv pre-compiled to ES5
I've been developing a React app using dotenv, with create-react-app, and when it came to building for production, the build failed due to some ES6 code in dotenv:

```
> react-scripts build

Creating an optimized production build...
Failed to compile.

Failed to minify the code from this file: 

 	./node_modules/dotenv/lib/main.js:23 

Read more here: http://bit.ly/2tRViJ9
```

I've been able to fix this myself by compiling dotenv to ES5 manually, but think it'd be a good idea for dotenv be published to npm pre-compiled to ES5, to prevent other people from running into similar issues in the future.

dawsbot:
Tests failed to begin with locally, so feel free to edit this PR to get those passing before merging. Seems like adding node 0.12+ to the tests was the best way to ensure this package is not using node 4+ only features


maxbeatty:
Node v0.12 reached end-of-life 2016-12-31 so we have stopped supporting it ([more](https://github.com/nodejs/Release)). If you need a version that works in that environment, try v4 and below.

PR #270, opened 2018-02-07
Parse objects inside envfile

ovictoraurelio:
Now is possible add a object inside .env like example below, that this object will be parsed

``OBJECT_EXAMPLE={``
``       "user": "example",``
``      #example comment``
``       "password": "teste"``
``}``


maxbeatty:
advanced parsing like this is better suited as a standalone module like the [ones listed in the README](https://github.com/motdotla/dotenv/blob/v5.0.0/README.md#go-well-with-dotenv)

PR #281, opened 2018-03-06
First PR

yTakkar:
Introduced `ES6` syntax to JS files.
And some minor changes in installation command and other projects' list of `README.md`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,11 +13,7 @@ Dotenv is a zero-dependency module that loads environment variables from a `.env
 ## Install
 
 ```bash
-# with npm
-npm i dotenv 
-
-# or with yarn
-yarn add dotenv
+npm install dotenv --save
 ```
 
 ## Usage
@@ -259,4 +255,3 @@ Here's some projects that expand on dotenv. Check them out.
 * [lookenv](https://github.com/RodrigoEspinosa/lookenv)
 * [run.env](https://www.npmjs.com/package/run.env)
 * [dotenv-webpack](https://github.com/mrsteele/dotenv-webpack)
-* [handy-browser-env](https://github.com/yTakkar/Handy-Browser-Env)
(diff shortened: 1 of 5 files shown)


zenflow:
This module currently supports node version 4, which does not support ES6 syntax (without the `--harmony` flag) so I think this can't be merged *right now*.

But node version 4 reaches End-Of-Life on 2018-04-30. So maybe then we can stop supporting node version 4 too, and switch to ES6 syntax.

zenflow:
@yTakkar I suggest breaking up this PR into separate PRs

yTakkar:
@zenflow I think you're right, I should break up this PR into separate PRs

zenflow:
@yTakkar Maybe open 3 fresh PRs rather than undoing changes here, so it's easier to review

maxbeatty:
👋 thanks for contributing! We've found more success in the past when someone creates an issue with a suggestion for a change; a majority of people agree on it, and then a Pull Request is created.

zenflow:
@maxbeatty Yeah, that approach is usually better, but for simple changes like f1cd386a5bc42e26433d9f5867756ddbeded3a94 it's easier and faster and makes more sense to suggest the change and discuss it in a PR, right?

zenflow:
I think the reason it's better to discuss non-trivial changes in an issue first is because it could be rejected and then your time coding the changes was wasted.

yTakkar:
But what if someone suggests any change and there's no issue about that change??

yTakkar:
I'll suggest merging it, after all it's a transition of code to es6 onyl!!

maxbeatty:
To minimize wasted effort, make sure your change is wanted before spending time on it. Issues are a good way to do that. Well-written Pull Request descriptions are another way. 

Today, I do not see enough benefit to accept these changes. If in the future new syntax results in a performance gain, safer execution, or similar improvement, I would be open to updating it.

zenflow:
> I mean PRs will never be accepted/merged if it doesn't have any related issue??

I think there are exceptions. See my comment https://github.com/motdotla/dotenv/pull/281#issuecomment-370867249. I don't know if @maxbeatty agrees because he didn't reply to me.

PR #286, opened 2018-03-10
Add rollup to keep compatibility and move forward

pedromig:
Allow the code to be transpiled so that new syntax may be used without sacrificing use in older versions of node.


maxbeatty:
What older versions of Node are you looking to support? 

We currently support 4.x (LTS) and above which was released in October 2015. I think we want to keep supporting LTS versions and nothing below to encourage the community to use supported versions of Node that still get security patches etc.

PR #288, opened 2018-03-12
Change in installation cmd & added a project to other projects list

yTakkar:


README.md
```diff
@@ -13,7 +13,11 @@ Dotenv is a zero-dependency module that loads environment variables from a `.env
 ## Install
 
 ```bash
-npm install dotenv --save
+# with npm
+npm i dotenv
```
maxbeatty: does `npm` save by default now? if not, should keep `--save` flag. I also have a strong preference for `install` over `i` for documentation for anyone not familiar with shorthand

zenflow: Yep, npm saves by default now. And yeah I agree with @maxbeatty that the explicit long-from command names are better for documentation

yTakkar: Yes npm saves to `package.json` by default now.

README.md
```diff
@@ -255,3 +259,4 @@ Here's some projects that expand on dotenv. Check them out.
 * [lookenv](https://github.com/RodrigoEspinosa/lookenv)
 * [run.env](https://www.npmjs.com/package/run.env)
 * [dotenv-webpack](https://github.com/mrsteele/dotenv-webpack)
+* [Handy-Browser-Env](https://github.com/yTakkar/Handy-Browser-Env)
```
maxbeatty: Can you explain how your module expands dotenv and encourages storing configuration in the environment separate from code?

yTakkar: When I was developing [React-Instagram-Clone-2.0](https://github.com/yTakkar/React-Instagram-Clone-2.0), I used `dotenv`, it worked like a charm on nodejs as I can access env variables very easily. But there was no package which allowed me to access those env variables on the front-side of the app, so I developed `Handy-Browser-Env`.

maxbeatty: A key distinction between server-side config and client-side config is that nothing sensitive can go into client-side config because you're distributing it to everyone (no matter how uglified it is, it's still being distributed). Config may still change between deploys for client-side apps but you're better off using JavaScript (or the language of your choice) directly to decide on that config, and it can be stored in git (or your vcs of choice) because, again, this configuration isn't sensitive.

I want to make sure this distinction is clear to developers. I see a lot of issues related to using dotenv in the browser and don't completely understand it. Adding a link to this module won't discourage using dotenv for browser-based projects. **If your module helps you, keep using it.** I'm going to request that this link be removed so that there's no added confusion.

yTakkar: Sure, I'll remove it if you want :)

PR #293, opened 2018-03-26
strip export statement to be compatible with other dotenv implementation

florisvink:
Dotenv autoloaders like [oh-my-zsh](https://github.com/robbyrussell/oh-my-zsh/tree/master/plugins/dotenv) needs the explicit shell export statement at the beginning of the line. This PR handles these .env files as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,6 +125,7 @@ The parsing engine currently supports the following rules:
 - `BASIC=basic` becomes `{BASIC: 'basic'}`
 - empty lines are skipped
 - lines beginning with `#` are treated as comments
+- export statements `export KEY=value` will be processed as it was `KEY=value`
 - empty values become empty strings (`EMPTY=` becomes `{EMPTY: ''}`)
 - single and double quoted values are escaped (`SINGLE_QUOTE='quoted'` becomes `{SINGLE_QUOTE: "quoted"}`)
 - new lines are expanded if in double quotes (`MULTILINE="new\nline"` becomes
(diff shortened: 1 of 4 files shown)


maxbeatty:
😕 the dotenv plugin you linked to says "export is optional"

This has been proposed a number of times and the thinking is still to support a consistent INI file format and encourage other modules that build on top of it from there

dosentmatter:
If anyone wants to allow exporting all variables in the `.env` file, you can do the following:

```
set -a # Turn on allexport
. ./.env # source .env
set +a # Turn off allexport
```

[`set -a` is POSIX compliant](http://pubs.opengroup.org/onlinepubs/009696799/utilities/set.html#tag_04_127_03) so it should work with most shells. Then you just don't include the optional export for all dotenv implementations.

eropple:
@maxbeatty "The thinking" may be this, but FWIW, coming from using sbt-dotenv and Ruby's `dotenv` and Rust's `dotenv` and literally every other `dotenv` package I've ever used, that's unexpected and, now, kind-of-problematic behavior because I can't use the same `.env` file for Node projects that I use for literally every other project for reasons that sound, to be a little bit blunt, dogmatic.

It's compounded by this package sitting on the name and making it really hard for any others to gain traction--there's a perfectly fine one called `nvar` that, frankly, is probably a superior `dotenv` to this `dotenv` just because _it doesn't act in surprising ways_. Indeed, the cause of the problem that led me here is that `nestjs-config` uses `dotenv` implicitly (because it has the mindshare) and so I can't fix it without forking `nestjs-config` and fighting an uphill battle.

The stance of this project actively makes working on Node more frustrating than it needs to be. I wish you'd reconsider.

PR #294, opened 2018-03-28
Convert bool/number to type

TheColorRed:
Converts numbers from strings to numbers and `true/false` strings to booleans.

Number examples:

* `NUMBER1=0.1 -> number`
* `NUMBER2=-12.4 -> number`
* `NUMBER2=-12.4.5 -> string`
* `NUMBER3=127.0.0.1 -> string`
* `NUMBER4=345 -> number`

Boolean examples:

* `BOOL1=true -> boolean`
* `BOOL2=FALSE -> boolean`
* `BOOL3=!true -> string`

I also added some basic get/set/add functions

```ini
ENV=dev
PORT=3000
HOST=127.0.0.1
```
```js
let env = require('dotenv').config()
console.log(env.get('ENV')) // returns dev
console.log(env.get('DB_HOST', '127.0.0.1')) // returns 127.0.0.1 since the key doesn't exist

// Attempts to add PORT but doesn't because it already exists
env.add('PORT', 1000)
console.log(env.get('PORT')) // returns 3000

// Updates PORT since it already exists
// If it didn't exist it will create it
env.set('PORT', 1000)
console.log(env.get('PORT')) // returns 1000
```

Lastly, I updated the doctype from `/*` to `/**` to make the doctype valid, as [mentioned](http://usejsdoc.org/about-getting-started.html#adding-documentation-comments-to-your-code):

> Each comment must start with a /\*\* sequence in order to be recognized by the JSDoc parser. Comments beginning with /\*, /\*\*\*, or more than 3 stars will be ignored.


maxbeatty:
Thanks for the idea but this was discussed a long time ago and now there are [modules](https://github.com/niftylettuce/dotenv-parse-variables) you can layer on top of dotenv if you want this

PR #298, opened 2018-04-12
Support encrypted .env.crypt files

shreeve:
Not sure if this is something that is wanted by others, but it's a helpful way to allow for `.env.crypt` files to be version controlled. In order to use these files, either set the `DOTENVCRYPT` environment variable with the decryption password or store it in a non-version-controlled file called `.env.cryptkey`. I'm not sure if this is of interest to others, or if it would be better to do this as a separate project or even a plugin to the main dotenv project, but I'm submitting it for review if there is interest.

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,6 +149,30 @@ control. It should only include environment-specific values such as database
 passwords or API keys. Your production database should have a different
 password than your development database.
 
+### How can I create a version-control-safe `.env` file?
+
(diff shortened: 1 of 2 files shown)


maxbeatty:
This would be better as a standalone module. We export `parse` to make this easy:

```js
const crypt = require("dotenv-crypt")

const config = dotenv.parse(crypt({ path: ".env.cryptkey" }))
```

PR #302, opened 2018-04-26
Drop node 4 support

### LINKED ISSUE #282: Drop support for node version 4 and start using ES6 syntax
Node version 4 reaches End-Of-Life on 2018-04-30 [(reference)](https://github.com/nodejs/Release#release-schedule)

What about stopping supporting node version 4 when it reaches EOL?

Related to #281

TheDancingCode:
Closes #282.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,10 @@
 language: node_js
 
 node_js:
-  - 4
   - 6
   - 8
   - 9
+  - 10
 
 jobs:
   include:
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -2,10 +2,10 @@ version: "{build}"
 build: off
 environment:
   matrix:
-    - nodejs_version: "4"
     - nodejs_version: "6"
     - nodejs_version: "8"
     - nodejs_version: "9"
+    - nodejs_version: "10"
 install:
   - ps: Install-Product node $env:nodejs_version
   - npm install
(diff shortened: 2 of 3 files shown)


maxbeatty:
Should [remove from appveyor](https://github.com/motdotla/dotenv/blob/1945a05028a1d7789745cef7f72f9bebaca4f492/appveyor.yml#L5) as well and add 10 to both CIs

TheDancingCode:
Tests on Node 10 are currently failing due to an issue with lab leak detection. It is fixed in v15 of lab, but upgrading is not an option as they now require Node >= 8.9. I've opened [an issue](https://github.com/hapijs/lab/issues/828) with them backport the fix to the v14 branch.

TheDancingCode:
Ava in turn [breaks when stubbing `fs`](https://github.com/avajs/ava/issues/1359) and other internals, which is exactly what happens in these tests. What about Jest?

maxbeatty:
Bummer. Ya we could try Jest or [node-tap](https://www.node-tap.org/). Minimal setup and coverage reporting are really the two big features we're looking for

zenflow:
For now maybe consider just disabling the feature of Lab that's causing the problem? (leak detection) Is it needed? I think most test frameworks (like these alternatives) don't have that feature anyways. Still would be nice to use a more reliable test framework when someone has time to do that work.

zenflow:
> Still would be nice to use a more reliable test framework when someone has time to do that work.

Ah like node-tap in #304

maxbeatty:
Closing this in favor of #304 would like to get Windows passing there then can cut a release

PR #303, opened 2018-04-26
Support `import.meta.url`-style path URLs

### LINKED ISSUE #301: Support `file://` URLs
`__dirname` is not available in ES modules, but it can be computed from `import.meta.url`. `import.meta.url` returns a fully qualified URL, including the `file://` scheme. When specifying a custom `path` option, it would be nice to be able to pass the URL in addition to regular paths. This way `__dirname` and `import.meta.url` can both be used with minimal processing at the call site.

ravinggenius:
Fixes #301

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -51,7 +51,7 @@ function config (options) {
 
   if (options) {
     if (options.path) {
-      dotenvPath = options.path
+      dotenvPath = options.path.replace(/^file:\/\//, '')
     }
     if (options.encoding) {
       encoding = options.encoding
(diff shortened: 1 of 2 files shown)


maxbeatty:
Sorry ES modules are not playing nicely out of the box, but I feel pretty strongly that `dotenv` should follow Node core by requiring a valid "`<string>` | `<Buffer>` | `<URL>` | `<integer>` filename or file descriptor" for [fs.readFileSync](https://nodejs.org/api/fs.html#fs_fs_readfilesync_path_options).

ravinggenius:
Understood. I opened an issue (https://github.com/nodejs/node/issues/20357) with Node, so hopefully it will be fixed there.

PR #304, opened 2018-05-02
drop support for Node v4

maxbeatty:
**BREAKING CHANGE** closes #282 

thanks to rework in #302 switched to node-tap for tests which gives us all the same features in ~30% less LOC

- [x] upgrade dependencies
- [x] remove examples (need to be better)
- [x] don't change anything meaningful in `lib/` (regex change was prompted by `standard`)

PR #306, opened 2018-05-04
Allow to use relative path in dotenv_config_path

vlaad360:
If file with absolute path is not found, try to use relative path to find env file

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -69,7 +69,7 @@ function config (options) {
   // try to use relative path instead
   try {
     const relativePath = path.resolve(process.cwd(), dotenvPath)
-    _returnParsedFile(relativePath, encoding)
+    return _returnParsedFile(relativePath, encoding)
   } catch (e) {
     return { error }
   }
--- a/lib/main.js
+++ b/lib/main.js
@@ -58,20 +58,38 @@ function config (options) {
     }
   }
 
+  let error = null
   try {
     // specifying an encoding returns a string instead of a buffer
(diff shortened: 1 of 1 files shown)


maxbeatty:
It's not clear to me how this would be used or why the current implementation is lacking

Not changing any tests and still getting 100% code coverage is also confusing me as to what is actually changing

vlaad360:
Sorry my bad
We just try to use `dotenv_config_path=/custom/path/to/your/env/vars`
And config not working, think that something possible about relative and absolute path's
But it's not

maxbeatty:
The [current CLI test for v5.0.1 uses a relative path](https://github.com/motdotla/dotenv/blob/v5.0.1/test/config.js#L16) fwiw. 

The [new proposed test](https://github.com/motdotla/dotenv/pull/304/files#diff-4564ba754e672af58eebf5252897cfa9R13) will not test this explicitly but the [option parsing logic](https://github.com/motdotla/dotenv/pull/304/files#diff-ee033622263ec12565b4d0a625fa7486R3) will be [tested independently](https://github.com/motdotla/dotenv/pull/304/files#diff-4f877f85bd8b8cc486856981feab29e3R7) which should make it easier to test if `/custom/path/to/your/env/vars` is being parsed properly. So [I added a test](https://github.com/motdotla/dotenv/pull/304/commits/8fc152cfd5df74be2524e557b64d6291507a1098) to do just that!

If you're still having trouble, it would make it easier to debug by making a simple project that we could use to reproduce

PR #313, opened 2018-05-30
config: simplify regex

caub:
explosive quantifiers can be nasty, that one was probably ok, but I think it's better to be more restrictive

### DIFF
--- a/config.js
+++ b/config.js
@@ -1,5 +1,5 @@
 (function () {
   require('./lib/main').config(
-    require('./lib/cli-options')(process.argv)
+    require('./lib/cli-options')(process.argv.slice(1)) // skip first argv, which is node command
   )
 })()
--- a/lib/cli-options.js
+++ b/lib/cli-options.js
@@ -1,11 +1,9 @@
-const re = /^dotenv_config_(.+)=(.+)/
-
-module.exports = function optionMatcher (args) {
-  return args.reduce(function (acc, cur) {
-    const matches = cur.match(re)
+module.exports = function (args) {
+  return args.reduce(function (options, val) {
+    var matches = val.match(/^dotenv_config_([^=]+)=(.+)$/)
     if (matches) {
-      acc[matches[1]] = matches[2]
+      options[matches[1]] = matches[2]
     }
-    return acc
+    return options
   }, {})
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -25,7 +25,7 @@
       "integrity": "sha1-r9+UiPsezvyDSPb7IvRk4ypYs2s=",
       "dev": true,
       "requires": {
-        "acorn": "3.3.0"
+        "acorn": "^3.0.4"
       },
       "dependencies": {
         "acorn": {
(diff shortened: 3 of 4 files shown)


zenflow:
This optimizes for performance over simplicity, no? Just pointing out that this doesn't "simplify" as the title says

caub:
@zenflow yes performance, but I realize it's not that much faster 

```js
var data = Array.from({length: 4e5}, () => `dotenv_config${Math.random()<.3?'_':'.'}foobar-qux${Math.floor(1e16*Math.random()).toString(36)}qu_ux_xu${Math.random()<.7?'=':'.'}${Math.floor(1e16*Math.random()).toString(36)}`)

var re1 = () => {
  data.forEach(val => {
    const m = val.match(/^dotenv_config_(.+)=(.+)/);
  });
};
var re2 =  () => {
  data.forEach(val => {
    const m = val.match(/^dotenv_config_([^=]+)=(.+)$/);
  })
};
var slice =  () => {
  data.forEach(val => {
    const m = val.slice(14).split('=', 2);
  })
};
bench(re1)
bench(re1)
bench(re2)
bench(re2)
bench(slice)
bench(slice)

bench(re1)
bench(re1)
bench(re2)
bench(re2)
bench(slice)
bench(slice)

function bench(fn) {
  console.time(fn.name);
  fn();
  console.timeEnd(fn.name);
}
```

maxbeatty:
Thanks for providing a benchmark to lay out the performance improvements. I'm not sure we should be optimizing for the use case of someone providing 400,000 command line arguments.

I created a test on jsPerf using [the newer option matcher](https://github.com/motdotla/dotenv/pull/304/files#diff-ee033622263ec12565b4d0a625fa7486) from #304 and what I expect most people to be feeding `process.argv`. Obviously a browser-based benchmark isn't apples to apples with a Node-based benchmark, but with V8 under the hood for both, I'm content.

https://jsperf.com/dotenv-cli-parsing/1

I'll leave this open for a bit so people can run the jsPerf test. I'm see the #304 implementation being ~40% faster. If that's the consensus, we can leave it as-is.

caub:
Here's a bench on node: https://repl.it/@caub/dotenv-argv-parse (not a huge fan of jsperf, accurate benchmarks can be done with console.time or perf_hooks)

config.js
```diff
@@ -1,7 +1,8 @@
 (function () {
   var options = {}
-  process.argv.forEach(function (val, idx, arr) {
-    var matches = val.match(/^dotenv_config_(.+)=(.+)/)
+  // skip first argb, which is node command
+  process.argv.slice(1).forEach(function (val) {
```
caub: again, not specially optimized for perf, it really doesn't matter at that scale, it's just more readable I think

zenflow:
I like this new version much better because it doesn't sacrifice any simplicity/readability

zenflow:
@caub thanks for the reference on explosive quantifiers!

maxbeatty:
Will you update this or open a new PR against `master` now that this code has moved?

lib/cli-options.js
```diff
@@ -1,11 +1,9 @@
-const re = /^dotenv_config_(.+)=(.+)/
```
maxbeatty: this PR could be simplified to:

```diff
-const re = /^dotenv_config_(.+)=(.+)/
+const re = /^dotenv_config_([^=]+)=(.+)$/
```

additional tests are always welcomed

maxbeatty:
closing in favor of #325

caub:
@maxbeatty ok, you might want that .slice(1) there https://github.com/motdotla/dotenv/blob/bfb2269ec4e0ec3794b8b68678eaa586cbfdaff0/config.js#L3 since first item is the node ccommand

PR #326, opened 2018-07-13
feat: sync config values from env vars

clakech:
We would like to get the config with overridden values from env variables.

Normal behavior of dotenv is to parse config and return parsed config without env vars

With this, you can use a syncWithEnv option to change config behavior and return parsed config WITH env vars override

PS: I will add test in this pull request if this feature is accepted

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,6 +109,16 @@ using this option.
 require('dotenv').config({encoding: 'base64'})
 ```
 
+#### Environment Variables Override
+
+Default: false
+
+You may specify the behavior of config to return parsed values with updated overridden environment variables values if found using this option.
+
+```js
+require('dotenv').config({syncWithEnv: true})
+```
+
 ## Parse
 
 The engine which parses the contents of your file containing environment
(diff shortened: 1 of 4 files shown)


maxbeatty:
Will you expand on why you can’t use `process.env` directly (which is the final source of truth)?

clakech:
I think process.env is nice to override config but to retrieve config it is slow and complexify tests

clakech:
@motdotla wdyt?

maxbeatty:
`process.env` should not be thought of as an override but rather the source of truth (i.e. "behave this way in this environment"). Reading from it should be just as performant as reading from any other object, and it is available anywhere which saves you from passing around an object wherever you need configuration.

`dotenv` is very good at reading the contents of a file and populating `process.env`. Reversing that logic (even as an option) is not something I think we should support at this time.

clakech:
Ok.

Reading from process.env is not as performant than reading an object. It is a getter reading your system variable with no cache. It is slower if you read it in each request and you have a lot of vars. You can check on the web or in the code of Node.

Using process.env everywhere in the code introduce a static dependency. You may have to mock process.env in your tests to be able to test a specific behavior. Your code is not pure anymore using process.env.

Your product, your decision by the way 😉

And last: using env for secrets may be unsecure even if 12factors recommand it https://diogomonica.com/2017/03/27/why-you-shouldnt-use-env-variables-for-secret-data/amp/?__twitter_impression=true

PR #331, opened 2018-08-20
add a new option `overwrite`

JounQin:
Enable to reload environment variables for nodemon, etc.


maxbeatty:
this introduces some potentially dangerous behavior (which has been discussed ad nauseam in other issues). if nodemon spawns a new process and your code in the process calls `dotenv`, you should realize changed variables

JounQin:
@maxbeatty But actually it can't be changed automatically with `nodemon`. Then I can only overwrite the environments with `dotenv` API, of course I can do it, but hope it could be better.

PR #334, opened 2018-09-07
added support for nested variables

morleytatro:
Not sure if there's interest, but the PHP variant of dotenv supports nesting variables. Could be helpful for concatenating values.

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,6 +132,7 @@ The parsing engine currently supports the following rules:
 - empty values become empty strings (`EMPTY=` becomes `{EMPTY: ''}`)
 - single and double quoted values are escaped (`SINGLE_QUOTE='quoted'` becomes `{SINGLE_QUOTE: "quoted"}`)
 - new lines are expanded if in double quotes (`MULTILINE="new\nline"` becomes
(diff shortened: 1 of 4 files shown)


morleytatro:
> You've placed your bullet above the multiline code block, it needs to be below so it doesn't sit between the multiline bullet and example 🙂

Thanks Tom! Fixed.

MurzNN:
I've faced this issue with different behavior of nested variables in php's `dotenv` and JS version. Here is an easy example of the problem - try to get the `MYSQL_ACCESS` from this `.env` file:
```
MYSQL_USER=user
MYSQL_PASSWORD=passw0rd
MYSQL_ACCESS=$MYSQL_USER:$MYSQL_PASSWORD
```
With PHP version it will be ok, with JS version - wrong.

 @maxbeatty can you please point me to the wrapper module, that resolves this problem?

MurzNN:
Seems you've talked about `dotenv-expand` - here is example with workaround:
```js
import dotenv from "dotenv";
import dotenvExpand from "dotenv-expand";

var myEnv = dotenv.config();
dotenvExpand(myEnv)
```

PR #335, opened 2018-09-14
Added link to env-path

benrei:


maxbeatty:
@benrei looks interesting and useful. we haven't discussed links in awhile and I have some evolving opinions

@motdotla @jcblw would the broader community be better served by just linking to who depends on the module and who mentions it as a keyword? I don't currently have time to endorse each project and don't want to play favorites

jcblw:
It would be nice to expose dotenv plugins in a more programmatic way. I am not quite sure of any solutions. If you have any @maxbeatty or @motdotla, I am all ears. That being said I think this PR is totally fine.

PR #336, opened 2018-09-26
fix: client side fails to run on IE11 if it requires/imports dotenv 

paulupendo:
This is happens on create-react-app on dev environment, specifically in IE11, as babel-loader is not set to transpile down any code to ES5 that lies outside the `src` folder. 

These changes will allow for seamless local dev processes on IE11 when using create-react-app and having dotenv as a dependency on client-side code.

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -58,7 +58,7 @@ function config (options) {
 
   try {
     // specifying an encoding returns a string instead of a buffer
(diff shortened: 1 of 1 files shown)


maxbeatty:
Sorry you're running into trouble. To be clear, `dotenv` should **not** be used in client-side code. It is only intended to be used in Node.js contexts.

PR #340, opened 2018-10-04
Allow configure dotenv via existing environment variables.

andrew-yustyk:
This is a very useful feature in a cases when .env file is located in different places on different servers 
or testing server has its own specific .env file commited to VCS etc.


maxbeatty:
We already allow this through the [path option](https://github.com/motdotla/dotenv/blob/v6.0.0/README.md#path), and you can easily use an environment variable to set it.

```js
require("dotenv").config({ path: process.env.DOTENV_CONFIG_PATH })
```

Issue #351, opened 2018-11-16
dotenv_config_path as environment variable itself [feature request]


damianobarbati:
Having to pass `dotenv_config_path` as process argument yields some downside:
- argv is polluted with it so the application has to handle this additional logic (filtering it out from args)
- concatenating command in a npm/yarn script may be messy in some cases when your command expects a second dynamic part (you have to deal with dotenv_config_path everywhere)

Is it possible to pass dotenv_config_path as an env variable itself?
Instead of:
```
$ node -r esm -r dotenv/config index.js dotenv_config_path=config/development.env
```
having:
```
$  DOTENV=config/development.env node -r esm -r dotenv/config index.js
```

damianobarbati:
@maxbeatty here the two scenarios I had trouble with and I couldn't come up with a solution.

1. let's assume I want to convert a command like:
```
A=1 B=2 C=3 D=3 AND=so-on npx -n='-r esm' jest --no-cache --detectOpenHandles"
```
 to dotenv.

- the following breaks npx -n flag:
```
npx -n='-r esm -r dotenv/config dotenv_config_path=x.env' jest --no-cache --detectOpenHandles"
```

- the following breaks jest because of unknown parameters
```
npx -n='-r esm -r dotenv/config' jest dotenv_config_path=x.env --no-cache --detectOpenHandles"
```

- the following would work seamlessly!
```
DOTENV=x npx -n='-r esm -r dotenv/config' jest --no-cache --detectOpenHandles"
```

This apply to any executables: expecting dotenv_config_path=value to be the first parameter of whatever the developer is working with/on is very obtrusive, while expecting a given ENV to be set is not and developer can opt-in dotenv without friction.

2. let assume I have a package script used as "base" for other scripts which have to load different environments. With an ENV a developer could easily do the following:
```
"run": "node -r dotenv/config my-script.js"
"run:dev": "DOTENV=dev.env yarn run myarg=1"
"run:stg": "DOTENV=stg.env yarn run myarg=2"
"run:prod": "DOTENV=prod.env yarn run myarg=3"
```

These are just my 2 cents on this, great package guys :)

abcd-ca:
@maxbeatty regarding [my pull request](https://github.com/motdotla/dotenv/pull/354#issuecomment-441783193) which I believe will also solve this issue, what more would you like to discuss?

maxbeatty:
So far, the reasons to support preload configuration using environment variables:

[...]

PR #354, opened 2018-11-23
Set dotenv_config_* vars via env var

abcd-ca:
Now `dotenv_config_*` can be supplied in a system env var instead of in CLI command.

Helps some IDEs that get confused with the CLI option like WebStorm/IntelliJ. WebStorm’s application of those command line config options was resulting in node trying to load them as additional npm modules rather than simple argv options that dotenv would consume.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,8 +65,8 @@ The configuration options below are supported as command line arguments in the f
 $ node -r dotenv/config your_script.js dotenv_config_path=/custom/path/to/your/env/vars
 ```
 
-Note, some IDEs like Webstorm and IntellJ might have trouble properly passing `dotenv_config_path` to node. This does not impact command line use, just when configuring run options within the IDE. 
-As an alternative for this situation, you can also set `dotenv_config_path` as an environment variable in your IDE run configuration instead of as a command line argument  
+If you are having trouble passing the `dotenv_config_<option>=value` part via the CLI, you can use environment variables instead. When using environment variables, be sure to use uppercase names, `DOTENV_CONFIG_<OPTION>=value`. 
+  
 
 ## Config
 
--- a/README.md
+++ b/README.md
@@ -65,6 +65,9 @@ The configuration options below are supported as command line arguments in the f
 $ node -r dotenv/config your_script.js dotenv_config_path=/custom/path/to/your/env/vars
 ```
 
+Note, some IDEs like Webstorm and IntellJ might have trouble properly passing `dotenv_config_path` to node. This does not impact command line use, just when configuring run options within the IDE. 
+As an alternative for this situation, you can also set `dotenv_config_path` as an environment variable in your IDE run configuration instead of as a command line argument  
+
 ## Config
 
 _Alias: `load`_
--- a/config.js
+++ b/config.js
@@ -2,6 +2,9 @@
 
 (function () {
   require('./lib/main').config(
-    require('./lib/cli-options')(process.argv)
+    {
+      ...require('./lib/cli-options')(process.argv),
+      ...require('./lib/env-options')()
+    }
   )
 })()
--- a/config.js
+++ b/config.js
@@ -2,9 +2,6 @@
 
 (function () {
   require('./lib/main').config(
-    {
-      ...require('./lib/cli-options')(process.argv),
-      ...require('./lib/env-options')()
-    }
+    Object.assign({}, require('./lib/env-options'), require('./lib/cli-options')(process.argv))
   )
 })()
--- a/config.js
+++ b/config.js
@@ -2,6 +2,6 @@
 
 (function () {
   require('./lib/main').config(
-    Object.assign({}, require('./lib/env-options'), require('./lib/cli-options')(process.argv))
+    Object.assign({}, require('./lib/env-options')(process.env), require('./lib/cli-options')(process.argv))
   )
 })()
--- a/lib/env-options.js
+++ b/lib/env-options.js
@@ -1,0 +1,13 @@
+/* @flow */
+
+const re = /^dotenv_config_(encoding|path|debug)$/
+
+module.exports = function optionMatcher () {
+  return Object.entries(process.env).reduce(function (acc, [key, val]) {
+    const matches = key.match(re)
+    if (matches) {
+      acc[matches[1]] = val
+    }
+    return acc
+  }, {})
+}
--- a/lib/env-options.js
+++ b/lib/env-options.js
@@ -1,7 +1,9 @@
 /* @flow */
 
-module.exports = {
-  encoding: process.env.dotenv_config_encoding,
-  path: process.env.dotenv_config_path,
-  debug: process.env.dotenv_config_debug
+module.exports = function (env = {}) {
+  return Object.assign({},
+    env.DOTENV_CONFIG_ENCODING ? { encoding: env.DOTENV_CONFIG_ENCODING } : null,
+    env.DOTENV_CONFIG_PATH ? { path: env.DOTENV_CONFIG_PATH } : null,
+    env.DOTENV_CONFIG_DEBUG ? { debug: env.DOTENV_CONFIG_DEBUG } : null
+  )
 }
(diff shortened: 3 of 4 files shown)


abcd-ca:
@motdotla My tests pass locally, I'm not sure what the problem is with the failed build here, I'm not understanding the reported problem.

abcd-ca:
@maxbeatty Thanks for taking a look! I'm not too sure if #351 is solving the same problem or not, it's hard to tell. What I know is that while passing the `dotenv_config_path` on the command line (bash) works because the variable is treated as an argv as it's supposed to. 

My IDE has some nice features that enhance mocha testing. As such, you can configure how to run the mocha tests from within the IDE. You have to tell it which node arguments to send, which environment variables, the working directory etc. The problem, is that when I put 

`-r dotenv/config your_script.js dotenv_config_path=/custom/path/to/your/env/vars`

as the node arguments, it tries to treat all of that above as part of the --require value. `dotenv_confi_path` does not end up in argv and I get an error that `dotenv_config_path` can not be found in node modules (it's trying to --require it). 

My fix allows dotenv to look in argv and then in environment variables. It allows me to then put the dotenv_config_path in the environment variables section in my IDE and I can then run my tests.

abcd-ca:
@maxbeatty I took a closer look at #351. He's trying to solve a different problem where variables meant for the dotenv library conflict with other apps that receive (argv) arguments that they don't expect and then complain about. I think he's saying that if variables destined for dotenv were supplied via environment variables (at least optionally), then they wouldn't conflict. I gather that in his case Jest is complaining about the unknown argument. 

In my case it's my IDE that isn't treating the dotenv variable as an argv variable, rather it's treating it as a node value of the --require flag; solved by use of environment variables. Note, I still like your original argv version – I prefer it and use it in my npm scripts but I want this workaround I created for situations where it's incompatible. I also included additional documentation to support this in the readme in this PR. Let me know if you have any more questions, I do think this fix would help with #351.

maxbeatty:
Both #351 and this PR are proposing the same addition: ability to configure dotenv via environment variables 

Let’s continue discussing if this is necessary in #351 and then if it is, come up with a plan to add

abcd-ca:
And here is the way that the IDE asks for the params. My fix works when I move the `dotenv_config_path` to the environment variables section now.
webstorm_dotenv_argv

abcd-ca:
@maxbeatty Can you help me see what the failed unit tests are complaining about? The tests pass for me locally

abcd-ca:
FYI, here's my local success with the same version of node that appveyor CI is dying at without much explanation:

```
myuser$ node -v
v6.14.4
myuser$ npm test

> dotenv@6.2.0-rc1 pretest /Users/myuser/Downloads/dotenv
> npm run lint


> dotenv@6.2.0-rc1 lint /Users/myuser/Downloads/dotenv
> standard


> dotenv@6.2.0-rc1 postlint /Users/myuser/Downloads/dotenv
> standard-markdown




> dotenv@6.2.0-rc1 test /Users/myuser/Downloads/dotenv
> tap tests/*.js --100

tests/test-cli-options.js ............................. 5/5
tests/test-config-cli.js .............................. 1/1 709ms
tests/test-config.js ................................ 14/14
tests/test-parse.js ................................. 17/17
total ............................................... 37/37

  37 passing (3s)

  ok
----------------------|----------|----------|----------|----------|-------------------|
File                  |  % Stmts | % Branch |  % Funcs |  % Lines | Uncovered Line #s |
----------------------|----------|----------|----------|----------|-------------------|
All files             |      100 |      100 |      100 |      100 |                   |
 lib                  |      100 |      100 |      100 |      100 |                   |
  cli-options.js      |      100 |      100 |      100 |      100 |                   |
  main.js             |      100 |      100 |      100 |      100 |                   |
 tests                |      100 |      100 |      100 |      100 |                   |
  test-cli-options.js |      100 |      100 |      100 |      100 |                   |
  test-config-cli.js  |      100 |      100 |      100 |      100 |                   |
  test-config.js      |      100 |      100 |      100 |      100 |                   |
  test-parse.js       |      100 |      100 |      100 |      100 |                   |
----------------------|----------|----------|----------|----------|-------------------|
```

config.js
```diff
@@ -2,6 +2,9 @@
 
 (function () {
   require('./lib/main').config(
-    require('./lib/cli-options')(process.argv)
+    {
+      ...require('./lib/cli-options')(process.argv),
+      ...require('./lib/env-options')()
```
maxbeatty: This syntax might be breaking Node 6?

Will want to flip order so CLI always takes precedence

lib/env-options.js
```diff
@@ -0,0 +1,13 @@
+/* @flow */
+
+const re = /^dotenv_config_(encoding|path|debug)$/
+
+module.exports = function optionMatcher () {
+  return Object.entries(process.env).reduce(function (acc, [key, val]) {
```
maxbeatty: Because we know exactly which 3 keys we want, this can be a simple object. May make testing easier to keep in a separate file

config.js
abcd-ca: Thanks @maxbeatty you're right, even though Node 6 supports es6 it does not support the spread operator on objects, only arrays. I'll update and flip precedence.

abcd-ca: strange that it worked when I tried it locally with same version of node though

lib/env-options.js
abcd-ca: I was trying to be consistent with the lib/cli-options.js which uses a reducer but it is simpler to just use a plain object.

config.js
abcd-ca: changes ready for review.

lib/env-options.js
abcd-ca: changes ready for review. Note I could change the case sensitivity [as mentioned here](https://github.com/motdotla/dotenv/issues/351#issuecomment-443817081) if you'd like.

lib/env-options.js
```diff
@@ -0,0 +1,7 @@
+/* @flow */
+
+module.exports = {
+  encoding: process.env.dotenv_config_encoding,
```
maxbeatty: Let’s use uppercase to match our existing documentation as well as Node’s official docs. Also, a simple unit test will help with integration tests.

README.md
```diff
@@ -65,6 +65,9 @@ The configuration options below are supported as command line arguments in the f
 $ node -r dotenv/config your_script.js dotenv_config_path=/custom/path/to/your/env/vars
 ```
 
+Note, some IDEs like Webstorm and IntellJ might have trouble properly passing `dotenv_config_path` to node. This does not impact command line use, just when configuring run options within the IDE.
```
maxbeatty: Let’s scrap this IDE-specific stuff and instead expand the existing docs to explain how to use new env vars. Examples outlined [here](
https://github.com/motdotla/dotenv/issues/351#issuecomment-443790144) would be a good start

config.js
```diff
@@ -2,6 +2,6 @@
 
 (function () {
   require('./lib/main').config(
-    require('./lib/cli-options')(process.argv)
+    Object.assign({}, require('./lib/env-options'), require('./lib/cli-options')(process.argv))
```
maxbeatty: Let’s add some integration tests to verify behavior that documentation outlines

README.md
abcd-ca: I think it would be helpful for those with IDE problems to mention something like, "If your IDE is having trouble passing the CLI config arguments like `dotenv_config_path` you can use environment variables instead (linked to that section)". It will save someone time, I'm sure. We don't have to mention any specific IDE this way. What do you think?

maxbeatty: It would be better to take IDE-specific questions to StackOverflow or other support forums. Write a blog post about how you fixed your issue. Those help people the most.

abcd-ca: OK, I'll do that and make the readme more generic

config.js
abcd-ca: added :)

maxbeatty:
to avoid more back and forth comments, I opened #355 based on this with what I had in mind

abcd-ca:
my pr works, your pr works

Issue #364, opened 2019-01-13
Custom parser


vranystepan:
Hello, 

first of all, thank you for all the amazing work you do on this library!
I'd like to ask whether is possible to add a custom parser to the `config` function?

**Background**
Sometimes, It's pretty challenging to pass some non-trivial values during the CI/CD process to the `.env` file. On the other hand, it's pretty easy to do that with JSON. Also, I get all secrets from Vault so everything's JSON :) 

**Proposal**
*Disclaimer:* I'm not developer, so please don't tear me to pieces :joy: 

```javascript
/*
type DotenvConfigOptions = {
  path?: string, // path to .env file
  encoding?: string, // encoding of .env file
  debug?: string, // turn on logging for debugging purposes
  parser?: function // custom parser
}
*/
```
...

```javascript
// Populates process.env from .env file
function config (options /*: ?DotenvConfigOptions */) /*: DotenvConfigOutput */ {
  let dotenvPath = path.resolve(process.cwd(), '.env')
  let encoding /*: string */ = 'utf8'
  let debug = false
  let parser = parse; // set default parser

  if (options) {
    if (options.path != null) {
      dotenvPath = options.path
    }
    if (options.encoding != null) {
      encoding = options.encoding
    }
    if (options.debug != null) {
      debug = true
    }
    if (options.parser != null) {
      parser = options.parser
    }
  }

  try {
    // specifying an encoding returns a string instead of a buffer
    const parsed = parser(fs.readFileSync(dotenvPath, { encoding }), { debug })
...
```

**Usage:**
```javascript
const dotenv = require('dotenv');

dotenv.config({
  path: 'env.json',
  parser: function(src, options) {
    return JSON.parse(src);
  }
});
```

maxbeatty:
Check out [dotenv-json](https://www.npmjs.com/package/dotenv-json) which does exactly this

PR #365, opened 2019-01-13
Custom parser

vranystepan:
As mentioned in #364, this minimalist adjustment allows to pass custom parser to the `config` function.  Then, dotenv will be able to handle cutom data types such as JSON. Default behavior of library **does not change**.


maxbeatty:
This is interesting but better off duplicating the file reading and object assignment code into purposeful modules like [dotenv-json](https://www.npmjs.com/package/dotenv-json) to allow sharing of custom parsers.

Issue #363, opened 2019-01-12
Base64 encoding not working?


sudhanshugaur4:
I am trying to use base64 encoding but I am getting key as undefined.

for e.g my actual string is 
> SECRET_KEY=hello1234

.env
U0VDUkVUX0tFWT1oZWxsbzEyMzQ=

process.env.SECRET_KEY// undefined

Can anyone please tell me is this a bug or I am doing something wrong?

maxbeatty:
dotenv will not decode the contents of your .env file

I'm assuming you are calling like this: `dotenv.config({ encoding: 'base64' })`. I've opened #366 to clarify the intended usage. Please reopen if this is not the case.

PR #366, opened 2019-01-13
clarify encoding usage

maxbeatty:
#363 likely arose from misleading documentation. updating docs and tests to use an encoding besides `utf8` that should still parse


jcblw:
Looks good! Thanks for the update.

PR #368, opened 2019-01-29
Update README.md

JasonSowers:
Added semicolon to  
```js
require('dotenv').config();
```


jcblw:
We use [Standard](https://github.com/standard/standard) as the style guide for this repo and we make the README consistent with that style. Thanks for PR but I think we are going to keep it as is.

PR #369, opened 2019-01-30
feat: 🎸($parse) improve the compatibility

zenboss:
improve the compatibility on the .env file line ending character

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -24,6 +24,8 @@ type DotenvConfigOutput = {
 const fs = require('fs')
 const path = require('path')
 
+const lineEnding = /\r?\n|\r(?!\n)/
+
 function log (message /*: string */) {
   console.log(`[dotenv][DEBUG] ${message}`)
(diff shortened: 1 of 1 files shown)


maxbeatty:
Is there a specific problem this is addressing? I would have expected to see an issue or two reported about incompatibility before seeing this change.

zenboss:
I have written a library named [readfileline](https://github.com/zenboss/readfileline) before. It had a problem when I use it on Windows OS, because the line ending in Windows is \r\n . And then, I followed the source code of Node.js and fixed it: https://github.com/nodejs/node/blob/master/lib/readline.js#L63

maxbeatty:
We aren't experiencing that problem with our Windows CI builds and haven't had any issues opened reporting problems. If either of those do happen, I'm happy to reopen this (probably with a test or two). For now, I prefer the simplicity and to follow "if it ain't broke, don't fix it"

PR #370, opened 2019-01-31
feat: support for overriding environment vars. fix #199

### LINKED ISSUE #199: Add an option which allows override environment variables
**It's a feature request.**

I think it would be cool if there will be option `override` or similar which can allow/disallow overriding of environment variables.
Like this:
```
require('dotenv').config({
  override: true
});
```
some discussions can be found here https://github.com/facebookincubator/create-react-app/issues/2230#issuecomment-302921036. 
What do you think?

userpixel:
This PR adds support for an `override` option to `config()` that is false by default.

The truth is that the workaround suggested [here](https://github.com/motdotla/dotenv#what-happens-to-environment-variables-that-were-already-set) is not the most elegant way of solving the problem and creates a messy [`setupfilesafterenv`](https://jestjs.io/docs/en/configuration.html#setupfilesafterenv-array) script in `jest`. This could easily be avoided if this feature was natively supported.
This feature is popularly demanded (https://github.com/motdotla/dotenv/issues/199) and can be great to add it.

I've updated the documentations with an example.
Also I've added tests with 100% coverage.

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,6 +117,23 @@ You may specify the encoding of your file containing environment variables.
 require('dotenv').config({ encoding: 'latin1' })
 ```
 
+#### Override
+
+Default: `false`
+
+When set to `true`, the environment variables from your env file override what is set in `process.env`.
+
+```txt
+# .env
+PORT=8080
+```
+
+```js
+process.env['PORT'] = '3000'
+require('dotenv').config({ override: true })
+console.log(process.env['PORT']) // prints 8080
+```
+
 #### Debug
 
 Default: `false`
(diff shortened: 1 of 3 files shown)


maxbeatty:
Please fork if you want this dangerous behavior

userpixel:
@maxbeatty why is it a dangerous behavior? At my job we have gone to great extent to work around it and that code is not less dangerous than having the feature in the dependency itself.

According to #199 it is a highly demanded feature (and it is opt-in which means people have to read the docs which prohibit this under normal circumstances).

maxbeatty:
I encourage you to also read through...

#34 
#55 
#63 
#72 
#123 
#130 
#137 
#231 
#331 

This isn’t a new request. I am firm on not supporting it. If you do not want to use the documented workaround, then the my next suggestion is to fork and maintain separately. I hope you can understand why I do not wish to spend any more time discussing.

userpixel:
Thanks for listing the relevant issues. I guess the discussion is over then.

---

I just leave this for anyone who wants to spends the time to contribute a PR:

The main counterargument is that you may accidentally override variables that screws up your code given three conditions:

1. You have environment variables with names like `PATH`, `PWD`, etc in your `.env` file
2. You explicitly choose the `override: true` in your `config()` options
3. Your code (or your dependencies) go crazy with values that you assign to them

`dotenv` chooses a safe default behavior which other packages [like this one](https://github.com/slowmove/dotenv-loader/blob/master/index.js) also use.

The unique selling point of `dotenv` module is to parse and set the environment variables. It is unfortunate that it doesn't complete the job but it gets you pretty far and you can always copy/paste [a workaround](https://github.com/motdotla/dotenv#what-happens-to-environment-variables-that-were-already-set) from its own site.

I'm not going to publish my fork just for this feature but the possibility to create a sister module that does exactly env var overriding is open.

PR #371, opened 2019-02-05
Add a callback to parse().

ericsnowcurrently:
There are a couple of use cases in which `dotenv.parse()` doesn't quite meet my needs:
* when I need to see the value from each valid line; sometimes an env var is set multiple times but currently I only get the last entry (I need them all)
* when the the order from the `.env` file matters

This change addresses those by adding a callback to `parse()` which gets called with the key/value pair for each valid env var line.  This could also be done with [an iterator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Iterators_and_Generators#Iterators) that gives the key/value pairs (and `parse()` would use that), but a callback seemed simpler.

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -29,7 +29,7 @@ function log (message /*: string */) {
 }
 
 // Parses src into an Object
-function parse (src /*: string | Buffer */, options /*: ?DotenvParseOptions */) /*: DotenvParseOutput */ {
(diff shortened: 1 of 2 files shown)


maxbeatty:
This use case is pretty far from the intended purpose of parse. Might be better off copying our code for your specific needs.

ericsnowcurrently:
Fair enough.  I'd already adapted your code locally but wanted to play nice. :)

PR #372, opened 2019-02-05
Be strict about env var names.

ericsnowcurrently:
Currently `dotenv.parse()` allows environment variable names which aren't actually valid for environment variables.  This change ensures that they are.


maxbeatty:
Can you elaborate on what makes for a valid environment variable? `process.env` is just an object and that’s what we are trying to support populating.

If we are going to update the battle-tested regex, I’d like to see a lot of tests.

ericsnowcurrently:
I was thinking of shells.  If folks are using .env files for more than actual environment variables then yeah a broader regex makes sense.  Regardless, I can filter bogus names after the fact.

Issue #357, opened 2018-12-06
Single white space after var in double quotes can invalidate variable


dpolicastro:
After editing my .env file using vi, my app broke and it took me some time to realize what happened;
I was declaring my env variables like that
```sh
MYSQL_USERNAME="myuser" <=
```
But unfortunately I added a white space before the last quote, like above, and when using `console.log()` to see which variables were being used in my config file, I got the following:
```js
...
 username: 'myuser"',
...
```

Issue #322, opened 2018-06-29
Trailing spaces in .env file cause parsing problems


jacbop:
## .env file
```
A="a value"
B="another value but with two trailing spaces"  
```

## .env characters
note the trailing whitespace after the double quote on the last line
```
$ od -a .env
0000000   A   =   "   a  sp   v   a   l   u   e   "  nl   B   =   "   a
0000020   n   o   t   h   e   r  sp   v   a   l   u   e  sp   b   u   t
0000040  sp   w   i   t   h  sp   t   w   o  sp   t   r   a   i   l   i
0000060   n   g  sp   s   p   a   c   e   s   "  sp  sp  nl
0000075
```

## index.js
```javascript
$ cat index.js
require('dotenv').config()
console.log(process.env.A)
console.log(process.env.B)
```

## output
```
node ./index.js
a value
another value but with two trailing spaces"
```

That trailing double quote should not be there

maxbeatty:
because you have trailing spaces after your quotes, [our regex](https://github.com/motdotla/dotenv/blob/v6.0.0/lib/main.js#L30) isn't removing the last quote (`['"]$`). additional logic could be added to account for that whitespace but I haven't fully thought through the pros and cons.

if you're looking for a quick workaround, make sure `"` is the last character on the line

fportela-ns:
if you do want the trailing spaces try setting the value without quotes, like:
`B=another value but with two trailing spaces  `

there's a [test](https://github.com/motdotla/dotenv/blob/master/tests/test-parse.js#L38) for this:
`t.equal(parsed.INCLUDE_SPACE, 'some spaced out string', 'retains spaces in string')`

PR #373, opened 2019-02-05
Make value capture is non-greedy.

ericsnowcurrently:
The value part of the regex in `dotenv.parse()` currently captures spaces.  This change adds an "?" (non-greedy modifier) to the regex to ensure it doesn't.

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -36,7 +36,7 @@ function parse (src /*: string | Buffer */, options /*: ?DotenvParseOptions */) 
   // convert Buffers before splitting into lines and processing
   src.toString().split('\n').forEach(function (line, idx) {
     // matching "KEY' and 'VAL' in 'KEY=VAL'
-    const keyValueArr = line.match(/^\s*([\w.-]+)\s*=\s*(.*)?\s*$/)
+    const keyValueArr = line.match(/^\s*([\w.-]+)\s*=\s*(.*?)?\s*$/)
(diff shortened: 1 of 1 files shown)


maxbeatty:
I can see how the current implementation allows for odd use-cases, but it is the most flexible way to allow consumers to do whatever they want (that's good from our perspective as it reduces support efforts).

_NB: for these examples, I'll use `.` to indicate a space_

If you want to include spaces in your values, declare a line such as `KEY=VAL...` and `process.env.KEY` will have a value of `VAL...`.

If you want to ensure spaces are trimmed for your values (because maybe you're dynamically generating them somehow or something), declare a line with quotes around your value such as `KEY="...VAL..."` and `process.env.KEY` will have a value of just `VAL`.

If you use quotes but include a trailing space, the closing quote will not be captured and the quote logic will not take place. This was the case in #357 and #322 and the reasonable fix was to remove the trailing space or not use quotes unless you need the additional logic (e.g. newlines). 

Putting the control in the consumer's hands allows them to do exactly what they want (even if it's greedy). Because this is not currently negatively impacting people, I'm going to keep it as-is ("ain't broke so won't fix").

ericsnowcurrently:
Got it.  Thanks for taking the time.  This was mostly a matter of me pushing upstream the fix I needed in order to use `dotenv.parse()` the way I expected it to work.

PR #374, opened 2019-02-05
Sort out quotation mark handling.

ericsnowcurrently:
Quotation mark handling in `dotenv.parse()` currently has a couple of small problems:
* leading marks without trailing marks (and vice versa) are still removed
* whitespace inside balanced (outer) quotation marks is removed, defeating part of the point of the quotation marks

This change corrects both.

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -44,14 +44,15 @@ function parse (src /*: string | Buffer */, options /*: ?DotenvParseOptions */) 
       // default undefined or missing values to empty string
       let value = keyValueArr[2] || ''
 
+      // remove any surrounding quotes
       // expand newlines in quoted values
       const len = value ? value.length : 0
       if (len > 0 && value.charAt(0) === '"' && value.charAt(len - 1) === '"') {
+        value = value.substring(1, len - 1)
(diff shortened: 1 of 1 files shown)


maxbeatty:
Why do you think removing whitespace from quoted values defeats part of the point of the quotation marks? Could the [last rule in our README](https://github.com/motdotla/dotenv/blob/406d0d1e380751e179703fb2056fb0f0511ee5d6/README.md#rules) be better?

ericsnowcurrently:
When I put quotation marks around a value I expect it to be treated as-is, including leading and trailing whitespace.  However, it sounds like that isn't how quotation marks are meant to be interpreted in `dotenv.parse()`, which is fine.  It just means that I can't use the library.

maxbeatty:
I can see how this is counterintuitive. You have to do less (no quotes) to get that behavior as outlined in #373 

The removing of unbalanced quotes does appear to be a problem and a tested fix would be welcomed.

ericsnowcurrently:
Unfortunately I won't have time (for now) to work on this further.  Thanks for taking the time.  Feel free to close this or leave it open, etc.

PR #378, opened 2019-02-19
ft: add a config option to toggle usage of process.env

adesege:
This PR adds a config option `returnProcess` to determine whether to use process.env or not.

This is suitable if you have a config module to abstract environment variables so as not to pollute the global scope with process.env and want it to work both in development and production


maxbeatty:
I'm sorry I don't understand the use case from your description. Our `config` method is intended to be used with the global `process.env`. If you do not want `process.env` populated, I would suggest using the `parse` method directly and using the returned values from there.

PR #385, opened 2019-04-04
Add parse value type feature

evsar3:
When parse an .env file into an object, parses values like `true`  or `false` or `123` into the correct type.
The way it is now, all values is parsed as strings.

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -60,7 +60,7 @@ function parse (src /*: string | Buffer */, options /*: ?DotenvParseOptions */) 
         }
       } else {
         // remove surrounding whitespace
-        val = val.trim()
+        val = parseValueType(val.trim())
       }
 
       obj[key] = val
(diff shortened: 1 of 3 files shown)


evsar3:
Seems that tests failed. I guess that's was already failling before this pull request. Going to take a look.

maxbeatty:
There are existing modules (e.g. [dotenv-parse-variables](https://www.npmjs.com/package/dotenv-parse-variables)) that handles this and that's the preferred way to extend dotenv (behavior that may not be universally beneficial).

I can't find the nodejs/node PR that undid implicit conversion but [the docs](https://nodejs.org/api/process.html#process_process_env) now read:

> Assigning a property on process.env will implicitly convert the value to a string. **This behavior is deprecated.** Future versions of Node.js may throw an error when the value is not a string, number, or boolean.

I think there's a future where this happens naturally 🤞

PR #386, opened 2019-04-05
Add ES import example in README.md

khaledosman:


README.md
```diff
@@ -27,6 +27,12 @@ As early as possible in your application, require and configure dotenv.
 ```javascript
 require('dotenv').config()
 ```
+or
```
akhilome: You could also specify that the `require('dotenv').config()` above is commonjs 🤔

maxbeatty:
There's already [a FAQ about `import` in the README](https://github.com/motdotla/dotenv/blob/v7.0.0/README.md#how-do-i-use-dotenv-with-import) as well as [an example](https://github.com/motdotla/dotenv/blob/v7.0.0/examples/typescript/src/lib/env.ts#L3) so I think we're' covered on this front. Thanks though!

PR #387, opened 2019-04-14
chore: add license badge

Hongarc:
Someone can see your license from `README.md`

PR #392, opened 2019-05-02
Drop support for Node v6

maxbeatty:
Dropping support for Node v6 as it has reached end-of-life.

This will be a **BREAKING CHANGE** as recommended by [nodejs/package-maintenance](https://github.com/nodejs/package-maintenance/blob/e2c519bc8a27bdb31d00066056fc2cf1bc8dc60b/docs/versioning.md)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,16 +7,16 @@ os:
   - windows
 
 node_js:
-  - "6"
   - "8"
   - "10"
   - "11"
+  - "12"
 
 cache: npm
 
 matrix:
   fast_finish: true
   allow_failures:
-  - os: windows
+    - os: windows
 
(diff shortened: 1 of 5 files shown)


maxbeatty:
Looking at [easy automatic nom publishes](https://blog.npmjs.org/post/184553141742/easy-automatic-npm-publishes) (and #387) lead me to [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0-beta.4/) and [standard-version](https://github.com/conventional-changelog/standard-version). Hoping they will make maintenance and releases easier.

Turns out AppVeyor isn't ready for v12 https://github.com/appveyor/ci/issues/2921

PR #393, opened 2019-05-05
Add option for multiple configurations based on NODE_ENV

soulofmischief:
With this option, dotenv can use multiple configurations by searching for `` `.env.${process.env.NODE_ENV}` ``.

This means instead of a single `.env` file, we can use `.env.production`, `.env.development`, etc.

It also works with custom paths.

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,6 +105,21 @@ You may specify a custom path if your file containing environment variables is l
 require('dotenv').config({ path: '/full/custom/path/to/your/env/vars' })
 ```
 
+#### MultiConfig
+
+Default: `path.resolve(process.cwd(), '.env')`
+
+Instead of loading `.env` or a custom path, load a configuration file with a suffix matching `NODE_ENV`.
+
+For example, if `NODE_ENV=production` then dotenv will look for `.env.production`.
+If `NODE_ENV=development` then dotenv will look for `.env.development`
+
+If using a custom path, this option will search for that path suffixed with `NODE_ENV`.
+
+```js
+require('dotenv').config({ multiConfig: true })
+```
+
 #### Encoding
 
 Default: `utf8`
--- a/lib/cli-options.js
+++ b/lib/cli-options.js
@@ -1,6 +1,6 @@
 /* @flow */
 
-const re = /^dotenv_config_(encoding|path|debug)=(.+)$/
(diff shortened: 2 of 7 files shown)


maxbeatty:
This has been discussed before, and I encourage you to read back through those old issues. I can't remember if someone created a module to handle this for you automatically, but the one-liner is:

```js

[".env", `.env.${process.env.NODE_ENV}`].map(path => dotenv.config({ path }))
```

soulofmischief:
@maxbeatty I did a search and nothing came up. Which issues can you think of?

I already have a one-liner in place which also accounts for an undefined NODE_ENV:

```
dotenv.config({
  path: path.resolve( paths.root,
    `.env.${ process.env.NODE_ENV || 'development' }`
  )
})
```
However I now have to place this *everywhere* I use dotenv, across projects, instead of a simple `dotenv({ multiConfig: true })`.

Why shouldn't this simple use case just be included in `dotenv`? I did all of the work already and it works fine. I just don't see any reason to not include the option when it greatly extends the usefulness of `dotenv` without any additional complexity. The fact that workarounds exist aren't a reason to not include the option.

maxbeatty:
Looks like [dotenv-flow](https://www.npmjs.com/package/dotenv-flow) does exactly what you want.

In the future when you'd like features added, I'd suggest starting with an issue where it can be discussed and gain support before spending time on implementation.

soulofmischief:
@maxbeatty `dotenv-flow` is a nice project but it is yet *another* dependency which aims to be much more complicated than this dirt simple implementation which covers 90% of use cases.

I would ask you to reconsider this PR by finding a good reason *not* to include it. Why is it better to turn users towards an unwieldy one-liner or plugin? Does the code smell? Are you not interested in extending the existing functionality of dotenv? If there's a good reason then there's no need to open an issue or discuss it any further. However you closed the PR without providing a reason.

maxbeatty:
> Make small focused modules for reusability and to make it possible to build larger more advanced things that are easier to reason about.
> – [Small Focused Modules](https://blog.sindresorhus.com/small-focused-modules-9238d977a92a)

dotenv does two things well: parsing INI-style files and assigning those parsed values to `process.env`. It’s well-tested, stable, and reliable. It’s also designed in a way to be easily extended and built upon. That is the preferred path forward for the reasons Sindre has outlined.

I encourage you to create your own module for your own needs.

PR #394, opened 2019-05-05
Add `test-html` script

soulofmischief:
Add a script which opens a coverage report in the browser.


maxbeatty:
`npm test -- --coverage-report=html` is another way to do this. Let's go with that pattern so we're not tempted to add a new npm script for each tap CLI arg

soulofmischief:
Got it, that's fine with me.

PR #395, opened 2019-05-05
Ignore coverage folder

soulofmischief:
https://github.com/motdotla/dotenv/pull/394 creates a directory named `coverage` in the project root.

This PR makes sure the folder is ignored.


maxbeatty:
Generating HTML coverage reports won't be normal behavior so leaving ignores as-is

PR #396, opened 2019-05-05
Add Node 12 import note

dandv:


README.md
```diff
@@ -253,7 +253,7 @@ errorReporter.client.report(new Error('faq example'))
 `client` will not be configured correctly because it was constructed before `dotenv.config()` was executed. There are (at least) 3 ways to make this work.
 
 1. Preload dotenv: `node --require dotenv/config index.js` (_Note: you do not need to `import` dotenv with this approach_)
-2. Import `dotenv/config` instead of `dotenv` (_Note: you do not need to call `dotenv.config()` and must pass options via the command line or environment variables with this approach_)
+2. Import `dotenv/config` instead of `dotenv` (_Note: you do not need to call `dotenv.config()` and must pass options via the command line or environment variables with this approach_). For Node 12, which mandates file extensions for `import`, use `import dotenv/config.js`.
```
maxbeatty: Does including the file extension work in all currently supported versions of node? If so, let's just update the initial instructions and any related example code to nudge people in that direction

dandv: I've tested it to work in Node 11.

maxbeatty: How about the other currently supported versions?

PR #399, opened 2019-05-22
Support for Prefix options

brianpunzalan:
Added another config options 'prefix' to support prefix check of keys. This is useful if you only need to get the keys matched with prefix. ex: CLIENT_APP_ prefix for client environment variables.


maxbeatty:
Thanks for the contribution. We're quite selective about new features and value discussion about the proposals. Community demand is also a strong driver (if something is only useful to a small group, we may not accept it).

Can you tell me more about your use case?

On the face of it, it seems like you should be able to create a file like `.env.client` with your `CLIENT_APP_` values and pass a custom path to dotenv.

brianpunzalan:
Yeah, I just thought that having an option `prefix` allows me to separate environment variables between my client and server. I had experience making separate `.env` files using path as mentioned, well it quite works but I was thinking maybe if the client and server env variables were stored in a single file `.env` and the prefix would be the the ones separates it then it would be nice.

I also noticed that `create-react-app` and `gatsby` have these prefixed env variables which makes me intrigued since its not within the `dotenv` option and they have their own implementation.

Another thing that comes to my mind is for every environments I have, I would need to maintain 2 files per environment.

maxbeatty:
Ah, any secrets sent to a client (e.g. browser) are no longer secret so we recommend just putting those configuration values in your client code directly. Build tools such as [webpack can help with this](https://webpack.js.org/plugins/environment-plugin/).

Closing this since the use case would provide a false sense of security.

brianpunzalan:
yes I agree on not presenting the secrets to client side in which I thought the prefixed variables for client are the ones I want to expose to the client side and the server could fetch all values or just the prefixed variables for server. 

I guess all client configuration could go to webpack configs, is much more preferred. Anyway, thank you for your suggestion.

PR #404, opened 2019-06-18
Support empty lines and comments.

wffranco:
In debug mode, show errors like:
```
[dotenv][DEBUG] did not match key and value when parsing line <n>:
[dotenv][DEBUG] did not match key and value when parsing line <n>: #<COMMENTED_LINE>
```
Sometimes we need to add empty lines to separate groups of variables, or comment some of them.
We don't want the debugger or the parser catch them.


lib/main.js
```diff
@@ -39,6 +39,11 @@ function parse (src /*: string | Buffer */, options /*: ?DotenvParseOptions */)
 
   // convert Buffers before splitting into lines and processing
   src.toString().split(NEWLINE).forEach(function (line, idx) {
+    line = line.trim()
+    // ignore empty and commented lines
```
maxbeatty: if you want to influence "debug" behavior, I'd suggest moving closer there instead of possibly altering the parsing logic. `if (debug && line.length && line[0] !== '#')` might do it, although I have reservations about _decreasing_ debug output.

maxbeatty:
Closing due to inactivity. Feel free to reopen but I would like to have more discussion about decreasing debug output.

motdotla:
@wffranco this was a good PR. It is a good use case. I merged it and modified it. Thank you for your contribution.

PR #413, opened 2019-08-13
Allow export in dotenv files

a666:
These are simple changes that allow leading export in dotenv environment variables.
Continues #118.
Should bring this dotenv repo in line with all the other implementations.

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,6 +172,7 @@ The parsing engine currently supports the following rules:
 {MULTILINE: 'new
 line'}
 ```
+- lines beginning with `export` are parsed normally
 
 ## FAQ
 
--- a/lib/main.js
+++ b/lib/main.js
@@ -29,7 +29,7 @@ function log (message /*: string */) {
 }
 
 const NEWLINE = '\n'
-const RE_INI_KEY_VAL = /^\s*([\w.-]+)\s*=\s*(.*)?\s*$/
+const RE_INI_KEY_VAL = /^\s*(?:export\s+)?([\w.-]+)\s*=\s*(.*)?\s*$/
 const RE_NEWLINES = /\\n/g
 
 // Parses src into an Object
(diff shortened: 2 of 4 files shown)


a666:
Completely unrelated, 
but in order to make the tests run I had to change the README.md file since the markdown linter was generating an error:

```
README.md
         204:10:   'k' is never reassigned. Use 'const' instead.
```

```diff
@@ -201,7 +201,7 @@ If you want to override `process.env` you can do something like this:
 const fs = require('fs')
 const dotenv = require('dotenv')
 const envConfig = dotenv.parse(fs.readFileSync('.env.override'))
-for (let k in envConfig) {
+for (const k in envConfig) {
   process.env[k] = envConfig[k]
 }
```

maxbeatty:
Thank you, but as stated in #118, there’s already a module that does this so it’s not needed here.

a666:
If I may say so, isn't that module redundant?

With every other dotenv file that we use for non js projects the developer can just do `source .env` and get to work. Now they would have to install a extra package and remember to use `source <(dotenv-export)`.

Besides, the essence of the PR has nothing to do with the programmatic sourcing of dotenv files, it just modifies the parsing of the file in order to make it less restrictive and more programmer friendly. And as a bonus, it also makes it less divergent with other implementations out there.

I understand this is a pretty popular project and major breaking changes could not be wanted, but this is, in the semver definition, a minor change that does not break backwards compatibility. 

If you believe it does then please let me know so I can test it further. 
If there is another issue, I'm completely open to ear it.

a666:
@maxbeatty Any further comment?

neoeinstein:
I was specifically looking for this feature. I understand your initial feelings on _not_ supporting `export`, but the convention for the file among the many implementations in various languages is to allow an optional `export` at the beginning of the line. Failing to do so was surprising to me, having come from other languages where that was supported.

PR #414, opened 2019-08-15
Parse \r line endings (mac)

### LINKED ISSUE #409: Can't get env ,when Line Separator is CR
Can't get env ,when Line Separator is CR

AlexanderGranhof:
Fixes #409 by finding all three EOL types (\n, \r, \r\n) and replacing it with the NEWLINE variable, currently being the Unix standard.

When i run `npm test` i get back `Error: test count exceeds plan`.

I have not used TAP to write tests before but i assume i need a plan to run the tests locally. However if i deliberately break the test it shows a different error, so im assuming the test works.

@maxbeatty

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -38,10 +38,8 @@ function parse (src /*: string | Buffer */, options /*: ?DotenvParseOptions */) 
   const debug = Boolean(options && options.debug)
   const obj = {}
 
-  // convert src to string and replace all EOLs to standard unix EOL
-  src = (Buffer.isBuffer(src) ? src.toString() : src).replace(NEWLINES_MATCH, NEWLINE)
-
-  src.split(NEWLINE).forEach(function (line, idx) {
+  // convert Buffers before splitting into lines and processing
+  src.toString().split(NEWLINES_MATCH).forEach(function (line, idx) {
     // matching "KEY' and 'VAL' in 'KEY=VAL'
     const keyValueArr = line.match(RE_INI_KEY_VAL)
     // matched?
--- a/lib/main.js
+++ b/lib/main.js
@@ -31,14 +31,17 @@ function log (message /*: string */) {
 const NEWLINE = '\n'
 const RE_INI_KEY_VAL = /^\s*([\w.-]+)\s*=\s*(.*)?\s*$/
 const RE_NEWLINES = /\\n/g
+const NEWLINES_MATCH = /\n|\r|\r\n/
 
 // Parses src into an Object
 function parse (src /*: string | Buffer */, options /*: ?DotenvParseOptions */) /*: DotenvParseOutput */ {
   const debug = Boolean(options && options.debug)
   const obj = {}
 
-  // convert Buffers before splitting into lines and processing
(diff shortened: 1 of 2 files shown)


lib/main.js
```diff
@@ -31,14 +31,17 @@ function log (message /*: string */) {
...
 const RE_NEWLINES = /\\n/g
+const NEWLINES_MATCH = /\n|\r|\r\n/
 
 // Parses src into an Object
 function parse (src /*: string | Buffer */, options /*: ?DotenvParseOptions */) /*: DotenvParseOutput */ {
   const debug = Boolean(options && options.debug)
   const obj = {}
 
-  // convert Buffers before splitting into lines and processing
-  src.toString().split(NEWLINE).forEach(function (line, idx) {
```
maxbeatty: `split` accepts a regular expression. did you try setting `NEWLINE` to your new `/\n|\r|\r\n/`? I'm hoping that's the only change needed 😺

tests/test-parse.js
```diff
@@ -58,6 +58,9 @@ t.equal(parsed.SPACED_KEY, 'parsed', 'parses keys and values surrounded by space
 const payload = dotenv.parse(Buffer.from('BUFFER=true'))
 t.equal(payload.BUFFER, 'true', 'should parse a buffer into an object')
 
+const macPayload = dotenv.parse(Buffer.from('MAC=true\rEOL=true'))
+t.same(macPayload, { MAC: 'true', EOL: 'true' }, 'can parse mac (\\r) line endings')
```
maxbeatty: Regarding your error, you need to update `t.plan` on line 12 to be 25 instead of 24 because you're adding this test. 

It would also be good to test for `\r\n` (from the regex). A combination of line endings might also be good to test just to cover our bases (but it's an unlikely real-world scenario).

lib/main.js
AlexanderGranhof: Oh, havent thought about that split accepts a regular expression. That would make it simpler. Adding it!

tests/test-parse.js
AlexanderGranhof: Ah i see, missed that t.plan. I'll add tests for all line endings.

lib/main.js
AlexanderGranhof: One thought i have is, if the src variable is used outside of split (it isn't currently) it could cause issues when not replacing all the EOLs with a standard one. Such as the `NEWLINE` variable.

But as it is currently it works fine.

maxbeatty:
released in v8.1.0

nmaxcom:
I wasted a long time debugging **my code** 😡 

image

AlexanderGranhof:
@nmaxcom 
Are you sure your using version 8.1.0 or later?
My pull request fixed your exact issue (i think?) with parsing \r line endings.

nmaxcom:
Nope. I created this project a few weeks ago and this is how it stands in my package.json: "dotenv": "^8.0.0"

I am grateful for this package, but I don't understand the low level of parsing up to version 8.1.0

PR #415, opened 2019-08-15
Added feature to enable watching of the configuration file and update

yusuf-khamis:
Auto-update of any configuration that has changed real-time without the need to restart app; unable to write tests for this particular feature though hence some tests failed. This watch will be activated upon request by configuration supply in config function as in some situations the app restart is required to successfully and smoothly apply the new configuration changes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,6 @@
 node_modules
 
 .DS_Store
+
+# Jetbrains config
+.idea
--- a/lib/main.js
+++ b/lib/main.js
@@ -77,6 +77,7 @@ function config (options /*: ?DotenvConfigOptions */) /*: DotenvConfigOutput */ 
   let dotenvPath = path.resolve(process.cwd(), '.env')
   let encoding /*: string */ = 'utf8'
   let debug = false
+  let watch = false
 
   if (options) {
     if (options.path != null) {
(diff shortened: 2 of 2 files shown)


maxbeatty:
This is a nifty idea, but I'm unsure about the practicality of it. How does the rest of the program learn about changes to `process.env`?

yusuf-khamis:
This library will update the `process.env` values on changes and the rest of the program will just pick them since its running on the same process. But as I have said, I have encountered this issue before that's why I came up with the idea.

maxbeatty:
Will you provide an example of this in action?

It has been my understanding that `process.env` is just a normal variable.

```js
var a = 1;
var b = a;
var a = 2;
// b is still 1
```

yusuf-khamis:
For example, if am consuming a third party API and want to update API keys, if I do this currently I have to restart my backend service to reflect the changes, but if I have the watcher then it's just real-time

yusuf-khamis:
All am saying it can be an option and user can decide whether or not to activate it

maxbeatty:
Thanks but I don't think this is going to provide enough usefulness within dotenv alone. It looks like there are two alternatives already published: [dynamic-dotenv](https://www.npmjs.com/package/dynamic-dotenv) and [dotenv-alive](https://www.npmjs.com/package/dotenv-alive)

PR #417, opened 2019-08-19
Added jsdocs for intellisense

AlexanderGranhof:
I added some basic jsdoc for the parse and config methods. This will add some neat intellisense for VS Code and possible other IDEs that support jsdoc.

image
image

I am not certain that the descriptions are the best, so any suggestions for better descriptions would be great.

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -18,8 +18,17 @@ type DotenvConfigOutput = {
   parsed?: DotenvParseOutput,
   error?: Error
 }
-
 */
+
+/**
+ * @typedef {Object} DotenvConfigOptions
+ * @property {String} [path] - Custom path to config file
+ * @property {"utf8" | "ascii" | "base64" | "binary" | "hex" | "ucs2" | "latin1"} [encoding] - Encoding to interpret file with. Default utf8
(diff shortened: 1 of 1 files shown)


maxbeatty:
We used to try to provide jsdocs but they became inaccurate over time. Is jsdoc useful outside an editor? Our API surface area is small enough that I think the README suffices so we wouldn't need to generate a site or anything based off of jsdoc.

If dotenv shipped its TypeScript types (instead of requiring installing `@types/dotenv`), would that provide proper intellisense?

From a maintenance standpoint, it'd be nice to only need to update README, TypeScript, and Flow. (Hope I'm not already forgetting something else)

AlexanderGranhof:
I see, understandable. I guess the best method would be to ship with an index.d.ts with all the types avaliable there. I have worked with Typescript before, but unsure if the type files work directly with a non-tsc compiled js file.

I could look into it and see whats possible with typescript and normal js files.

maxbeatty:
If you’re up for it:

1. Bring `@types/dotenv` here (and deprecate those)
2. Ship TS types as 8.2.0 (I can help with release)
3. Verify VS Code shows intellisense based on TS even in JS projects

If that doesn’t work as expected, we can revisit jsdocs.

PR #425, opened 2019-10-03
DotenvConfigOptions: new parameter

innomaxx:
Added optional `overwrite` parameter to override upper-level-defined environmental variables with values from `.env`


maxbeatty:
Please look at previously closed issues and pull requests about this behavior. There are other modules that will allow you to do this. We won't be adding it here.

PR #432, opened 2019-10-17
change resolved path to be absolute to the current project activate d…

Adidi:
If you activate your project using dotenv from different location for example you are running your node project as a global cli then `dotenv` won't load configs properly since its using process.cwd() as the resolved path: https://github.com/motdotla/dotenv/blob/70425a9c88e5fe5c3bd128fa973701279a76a9e3/lib/main.js#L78

changing it to `__dirname` make it absolute to the project root thus working from everywhere

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -75,7 +75,7 @@ function parse (src /*: string | Buffer */, options /*: ?DotenvParseOptions */) 
 
 // Populates process.env from .env file
 function config (options /*: ?DotenvConfigOptions */) /*: DotenvConfigOutput */ {
-  let dotenvPath = path.resolve(process.cwd(), '.env')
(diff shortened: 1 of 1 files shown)


maxbeatty:
`__dirname` will resolve to `your_project/node_modules/dotenv/lib` because it is "the directory name of the current module" which will always be "dotenv" wherever it is nested in your dependencies.

In your application code, you should use `__dirname` to configure dotenv if you are starting  your project from different locations.

```js
const path = require("path")

const { error } = dotenv.config({ path: path.resolve(__dirname, ".env") } )

if (error) { console.error(error) }
```

Adidi:
Got you!
Thanks :)

PR #440, opened 2019-11-20
fixes #439

### LINKED ISSUE #439: Escape of shell special characters does not work properly
*Given*
```sh
cat .env
FOO=Dollar\$Sign
```
*when*
```sh
source .env
```
*then*
```sh
echo $FOO
Dollar$Sign
```

when you do the same with dotenv then this behaves different:

```javascript
const { config } = require('dotenv');
let vars = config({path: './.env'});
console.log(vars);
{ parsed: { FOO: 'Dollar\\$Sign' } }
```
as you can see `\$` is kept, so the escaping of `$` is not recognised correctly. 
That makes working with passwords quite tricky.

sassman:
there are a couple of differences when it comes to how dotenv treats escaping characters `\` for characters were the shell would require escaping. One example is given in #439 

In fact some passwords have some strange characters, so If I want to be able to source my .env file (for production) and use dotenv (for tests) then I cannot do that with one single .env file. 

So here I propose a (possibly breaking) change that would support a bit of the behaviour that the shell has regarding escaping.

I appreciate any sorts of feedback or thoughts on this topic.


sassman:
travis failed on the windows build node by random git clone error. I guess retrigger the build will help.

sassman:
@maxbeatty any feedback is very welcome

blbradley:
I'm experiencing this with a password with dollar signs in it.

blbradley:
I tried your patch but it doesn't work for me.

blbradley:
I used npm-shrinkwrap to override the version of dotenv used with serverless-dotenv.

blbradley:
I tried all variations of with and without single quotes and backslashes.

mutantcornholio:
@motdotla @jcblw any chance to land this?

motdotla:
conversation for this moved here: https://github.com/motdotla/dotenv/issues/521

I think we are likely to move this direction but it will be a breaking change that we have to be quite careful of.

Issue #438, opened 2019-11-14
Path option documentation is unclear


tiendq:
Example of `path` option is unclear that it should has a file name at the end or not. The path value isn't so helpful as author might expected.

`require('dotenv').config({ path: '/full/custom/path/to/your/env/vars' })`

I think a simple example like this is crystal clear:

`require('dotenv').config({ path: '/var/www/myapp/.env' })`

drtobbyas:
Also, Can I use relative path instead of absolute path

jcblw:
There is so much deviation between file system structures between computers and operating systems I am not sure there is a one path that fits all. @tiendq  I am not sure if your example is any clearer to me then the one in the README.

Even [node.js documentation](https://nodejs.org/dist/latest-v12.x/docs/api/fs.html) uses this method of showing file paths in its documentation.

```javascript
const fs = require('fs');

fs.open('/open/some/file.txt', 'r', (err, fd) => {
  if (err) throw err;
  fs.close(fd, (err) => {
    if (err) throw err;
  });
});
```

@drtobbyas yes you can, but you should probably use something like path resolve to make sure it consistent.

eg.
```javascript
path.resolve(process.cwd(), './.env')
path.resolve(__dirname, '../../.env')
```

tiendq:
@jcblw You just missed my point. Are you sure they are same? `file.txt` is definitely a file name while `vars` is a vague name.

`'/full/custom/path/to/your/env/vars'`

`'/open/some/file.txt'`

jcblw:
@tiendq I assuming the issue is because `.env` is not in the path even though that is not a requirement of the option. This option is often used to load different environment files like `.env.development` or some other env file. I would not want to give the perception that this has to be a file named `.env`.

tiendq:
@jcblw It does need a file name in the path, the example doesn't have one, this is the issue.

jcblw:
@tiendq yes understandable, but the file does not need to have .env in the name. In the example vars could be a file on the example. 

Talked it over @motdotla last night. We are going to make an update with `.env` in the path just so the section is a bit clearer.

PR #441, opened 2019-12-05
README: clarify that path is to a filename and not a folder

motdotla:
https://github.com/motdotla/dotenv/issues/438


motdotla:
@tiendq does this help?

tiendq:
> @tiendq does this help?

Sure, it does, thanks a lot.

Issue #437, opened 2019-10-29
🔐👻 Multiple .env files, encrypting secrets, and committing .env to code


lorensr:
I'm thinking it would be helpful to many for this section of the docs to be expanded/clarified.

https://github.com/motdotla/dotenv#should-i-have-multiple-env-files

![image](https://user-images.githubusercontent.com/251288/67774960-a3829c00-fa34-11e9-8cd6-f4d12d9ac845.png)

Let's say I have a `.env` file with development values, and it's working as I'm developing with `npm run dev`. Now I want to do `npm run deploy`, but the `.env` values need to be different in production. How do I not have a separate `.env` file? By editing my single `.env` file every time I deploy? 😄

GeorgeTaveras1231:
@lorensr so funny. I was just reading the docs and was just as confused by this section, and was going to open a similar issue 😄.

My interpretation of what the author meant is that you should avoid having a hierarchy of env files that are combined for a single "deploy".  The message seems to be to maintain a clear separation between configs used for each "deploy".

Whether you do this by maintaining multiple files in your repo such as `.env.development`, `.env.staging`, etc. OR by having a single `.env` file that gets added to the project during the deployment process, I think should be left up to the developer.

On a project I'm currently working on, I've chosen to maintain all of the env vars in the repo, to make it easier to maintain and get the benefits of version control. Given that there are secrets in my env vars, I've chosen to have two env files for each deployment environment, one `public.env` (for env vars that don't have sensitive information (like `UNION_DEPLOYMENT_ENV` -- and env var that tells our internal libraries which deployment environment to target when referencing shared assets), and a `private.env` which is encrypted using a secret key we keep out of the repo.

So my project looks something like

```
/config
  /development
      /public.env
      /private.env.asc
  /qa
      /public.env
      /private.env.asc
  /production
      /public.env
      /private.env.asc
```

I'm not familiar with the 12 factor app, but I'd be curious to hear more about wether or not this approach is compatible with it's philosophies, and why.

giacomocerquone:
Well, we can absolutely say that it is a confusing section of the readme. I was about to open the same issue and yes, it meant what @GeorgeTaveras1231 said.
`you should not be sharing values between environments`, so it's saying that even if you have a variable with an equal value in env.prod and env.stage, that must be defined two times, each in every single file because you'll inject just one of those two files into your build.

[...]

PR #444, opened 2019-12-17
Safe with defaults

frontsideair:
In this PR I introduce handling of two extra files, `.env.default` and `.env.example`.

`.env.default` allows us to have default values. This should be used only for sane defaults, like the port number for a database. Usernames or passwords should never be put in there.

`.env.example` has two purposes: First, it allows the user to copy it as `.env` and modify it as a boilerplate. Second, it will throw if any of the variables is not provided. This accounts for defaults, `.env` and explicitly passed environment variables.

This behavior does not alter the existing functionality and it can be turned out by opting-in using the `extended` parameter. Default and example filenames can be overridden as well.

List of additional environment variables used:
- `DOTENV_CONFIG_EXAMPLE_PATH`
- `DOTENV_CONFIG_DEFAULT_PATH`
- `DOTENV_CONFIG_EXTENDED`

List of additional command line arguments:
- `dotenv_config_example_path`
- `dotenv_config_default_path`
- `dotenv_config_extended`

List of additional parameters to `config`:
- `examplePath`
- `defaultPath`
- `extended`

I have refactored the code a bit in the first commit to implement the features more easily. Everything introduced here is tested and 100% test coverage is maintained.

I'd really like for this to be merged and I can work on the potential issues. This is not complete but it's in a good state for reviews.

Once the API design is solidified, I can work on documenting the changes.

### DIFF
--- a/lib/cli-options.js
+++ b/lib/cli-options.js
@@ -1,12 +1,21 @@
 /* @flow */
 
-const re = /^dotenv_config_(encoding|path|debug)=(.+)$/
+const re = /^dotenv_config_(encoding|path|debug|example_path|default_path|extended)=(.+)$/
+
+const map = {
+  encoding: 'encoding',
+  path: 'path',
+  debug: 'debug',
+  example_path: 'examplePath',
+  default_path: 'defaultPath',
+  extended: 'extended'
+}
 
 module.exports = function optionMatcher (args /*: Array<string> */) {
   return args.reduce(function (acc, cur) {
     const matches = cur.match(re)
     if (matches) {
-      acc[matches[1]] = matches[2]
+      acc[map[matches[1]]] = matches[2]
     }
     return acc
   }, {})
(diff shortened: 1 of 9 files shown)


jhanzo:
Very nice projects, I am eager to see it released ! Any maintainer updates on this PR ?

jcblw:
Hey there @frontsideair, just looking at the PR.

I know we are not really looking to support multiple files. We even [advocate against it](https://github.com/motdotla/dotenv#should-i-have-multiple-env-files), but it seems like there is some debating going on with that [specific section](https://github.com/motdotla/dotenv/issues/437) of our readme.

I was just wondering what is the use case for this? Can this functionality be added by a plugin?

I do see a plugin about dotenv defaults.

https://www.npmjs.com/package/dotenv-defaults

jcblw:
Closing PR since this is probably not a feature we would want in the core of `dotenv` and the lack of a response. @frontsideair  if you still think there is a valid reason for this, please just respond on this thread.

frontsideair:
Hi, thanks for looking at this and detailed response! I had some arguments against the choice of single file but since I don't have time anymore to work on this maybe it can wait. I still have my fork that I can use for my own purposes. I can try to find time to work on this in case it gets revived in the future though!

Issue #448, opened 2020-01-08
TypeScript - Dependency Injection Type


PizzaPokerGuy:
**Objective**: have a type for the full dotenv package when being used during dependency injection. Example is from a nestjs/node project.

Current Workaround:

`
constructor(private dotenvModule: typeof dotenv = dotenv) {}
`

Preferred: 
`consructor(private dotenvModule: dotenv) {}`

jcblw:
Open to a PR for this, seems like it would only be making a interface for the dotenv that includes: load, config, and parse function types.  Here is the [type definitions](https://github.com/motdotla/dotenv/blob/master/types/index.d.ts)

motdotla:
We will come back to this soon. Moved here: https://github.com/motdotla/dotenv/issues/518

PR #450, opened 2020-01-30
Add main interface for library

MetCoder95:
As requested in the following Issue: #448 

The PR exports a main interface for the dotenv library.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -57,7 +57,7 @@ export interface DotenvConfigOutput {
 export function config(options?: DotenvConfigOptions): DotenvConfigOutput;
 
 /** dotenv library interface */
-export interface dotenv {
+export interface DotEnv {
   config: typeof config;
   parse: typeof parse;
 }
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -55,5 +55,12 @@ export interface DotenvConfigOutput {
  *
  */
 export function config(options?: DotenvConfigOptions): DotenvConfigOutput;
+
+/** dotenv library interface */
+export interface dotenv {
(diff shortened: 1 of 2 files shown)


jcblw:
@MetCoder95 looks great and thanks for the PR 🎉

types/index.d.ts
```diff
@@ -55,5 +55,12 @@ export interface DotenvConfigOutput {
  *
  */
 export function config(options?: DotenvConfigOptions): DotenvConfigOutput;
+
+/** dotenv library interface */
+export interface dotenv {
```
jcblw: Can we export out this interface with the casing `DotEnv`? It would then match the casing of the other interface [here](https://github.com/motdotla/dotenv/pull/450/files#diff-88bec6beae84369c0376b56b3bb88fe1L44).

jcblw: actually I think i am wrong about this let me download the code and get back to you.

jcblw: Ok pulled down, a saw the behavior of this. I think we should Pascal case the interfaces. I would love for the typing to work with. 

```
import dotenv from 'dotenv'

dotenv.foo() // throws on compile.
dotenv.config() // works 
```

I wish I was better at typescript and could add some code of how to do this, but sadly I am not quite sure how to get this to work.

MetCoder95: To do that, we can use [namespace](https://www.typescriptlang.org/docs/handbook/namespaces.html) to make it 👍

MetCoder95:
Done! 👍 
Sorry for the delay, I was quite busy.
Let me know if you need more changes or comments

PR #460, opened 2020-03-23
fixex #457

### LINKED ISSUE #457: did not match key and value for npm registry
.env
`npm_config_@myname:registry=https://npm.myname.com`

[dotenv][DEBUG] did not match key and value when parsing line 13: npm_config_@myname:registry=https://npm.myname.com

simplymichael:
Fix issue-#457: 'did not match key and value for npm registry' by adding support for the at sign (@) and colon (:) in variable names

fixes #457

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -29,7 +29,7 @@ function log (message /*: string */) {
 }
 
 const NEWLINE = '\n'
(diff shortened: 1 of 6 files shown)


motdotla:
thanks for this PR simplymichael. clever solution, but we're going to avoid supporting @ sign and colon in variable names for now. it's a practice i think we should avoid as a community.

but maybe others can weigh in here and leave their thoughts on why it is useful?

PR #473, opened 2020-05-24
fix: debug option

daychongyang:
### When I configure the `debug` option to `false`, __it cannot run correctly__, and the type check can only be set to boolean or undefined,  which would be very unfriendly.

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -86,7 +86,7 @@ function config (options /*: ?DotenvConfigOptions */) /*: DotenvConfigOutput */ 
     if (options.encoding != null) {
       encoding = options.encoding
     }
-    if (options.debug != null) {
+    if (options.debug) {
       debug = true
     }
   }


OmgImAlexis:
@motdotla what's holding this up from being merged? If it's the tests can they please be re-run as this shouldn't be the cause of the failure.

motdotla:
this is going out tonight from here: https://github.com/motdotla/dotenv/pull/519

Thank you so much for this and big apologies for the huge delay.

motdotla:
so this seems a bit unwieldy or unexpected but we are going to stick with the same behavior for now. 

the `debug` option here is actually intended to be either unset or set. https://github.com/motdotla/dotenv/pull/473/files#diff-7934bf411fea192ad8cd69e0a12911648a2842cb0f2409a8fb67b41b7069d757L14

daychongyang:
Oka.51 joy!

OmgImAlexis:
@motdotla any specific reasoning for this? Seems like an odd choice.

motdotla:
not a strong reason at this point. wanted to keep minimal changes at this point as mentioned above. in the near future, we will likely readdress this.

Issue #475, opened 2020-07-06
VS Code cannot find "dotenv/types"


alexesprit:
**TL; DR:** VS Code displays an error it can't find `dotenv/types`, this can be fixed by replacing the `types` property of the `package.json` file with "types/index.d.ts".

I have a project written in JavaScript, where `dotenv` is used as a dev dependency. I added a `jsconfig.json` file to the project recently, and noticed an error VS Code displayed for this file:
```
File '%path_to_project%/node_modules/dotenv/types' not found.
```
<details>
  <summary>Screenshot (Click to expand)</summary>

Content
![image](https://user-images.githubusercontent.com/1119267/86647333-a2a10180-bfe8-11ea-8192-87adf082b054.png)
</details>

I found out the `types` property of dotenv's `package.json` file is set to "types" which is directory where `index.d.ts` is stored. For some reason VS Code `ts` checker treats "types" as a file, and cannot find it, as it does not exist.

I changed this property to "types/index.d.ts" in the `package.json` right in my `node_modules` directory. I see no errors in VS Code after this change. If this fix is acceptable I can make a PR with this change.

Also, in TypeScript documentation is said the `types` property should point to a declaration file, not a directory.
Ref https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html#including-declarations-in-your-npm-package.

Interesting that seems like VS Code is able to load the `index.d.ts` file in some cases (e.g. it displays documentation and types for dotenv functions in the code editor), even if I keep `types` property untouched. Not sure why it does work in the editor, and why it displays error for the `jsconfig.json` file at the same time.

zenflow:
@alexesprit Yeah, the "types" property in package.json is a file (not a directory) in every project I've ever seen. 

I would go ahead and make a one-liner PR if I was you. Cheers!

riyadshauk:
This fix doesn't seem to work for me... Did I miss something? I'm just doing the standard `// @ts-ignore` over my import of `dotenv` for now. Conveniently, this isn't too problematic for me since I don't really use `dotenv` in my code much, anyways.

alexesprit:
@riyadshauk not sure why it does not work for you. I guess you restarted VS Code after applying changes, or closed it before applying changes.

By the way, thank you for the `ts-ignore` tip! I'll try this way.

Anoesj:
Restarting VSCode fixed this, thanks. Weird bug though..

pierlucg-xs:
[...]

PR #476, opened 2020-07-08
fix(package.json) : add the exact types file instead of the directory

axel-andre:
This PR is here to solved the issue #475. This is an quick and easy fix at first proposed by @alexesprit.

This problem is actually pretty annoying when you are using VSCode. By pointing to `index.d.ts` directly we avoid the issue and it does not make any breaking change.

Again, pretty simple PR tho.

(First contribution ever, do not be afraid to tell me if anything is wrong !)

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "version": "8.2.0",
   "description": "Loads environment variables from .env file",
   "main": "lib/main.js",
-  "types": "types",
+  "types": "types/index.d.ts",
   "scripts": {
     "flow": "flow",
     "dtslint": "dtslint types",


OmgImAlexis:
@motdotla what's holding this up from being merged? If it's the tests can they please be re-run as this shouldn't be the cause of the failure.

Anoesj:
Please merge this PR 🙃

OmgImAlexis:
@motdotla can you please check on this issue. This would help a fair few people and it doesn’t break anything.

Axel-Andre:
The CI error is not caused by my single line edit. Is it possible to merge this ? Or at least answer here @motdotla when you have time

SeedBoot:
While not an exact solution, I tend to use dotenv as part of a `package.json` script e.g. `nodemon -r dotenv/config`. No errors so far 🤷

sirlancelot:
The bug described in #475 is affecting my project only when I open the file `jsconfig.json` in VSCode. My project doesn't directly depend on `dotenv`, it is included by one of my dependencies (`@vue/cli-service`).

I can confirm that applying this patch (setting types to the exact file instead of the directory) fixed the error.

According to TypeScript's official documentation for [publishing](https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html), the `types` property must point to a file.

OmgImAlexis:
@motdotla can you please check on this issue. It’s been a while since we last pinged you.

motdotla:
this will be released here: https://github.com/motdotla/dotenv/pull/519

Issue #391, opened 2019-04-26
Please implement simple multi-line values


jorgeorpinel:
I.e. reopen https://github.com/motdotla/dotenv/issues/28 please

Should match the Ruby functionality: https://github.com/bkeepers/dotenv#multi-line-values

Thanks

maxbeatty:
Please do not open issues just to ask others to be reopened or reconsidered.

We have no intention of matching functionality (this has been the stance for a very, very long time).

`dotenv` is very stable. The use cases for this feature have not be stated well. Adding this feature would require substantial changes to `parse` which is high risk. High risk changes to stable code for poorly understood use cases is not a recipe for success.

jorgeorpinel:
Boo

Issue #28, opened 2014-06-22
Multi-line variables with double quotes?


randallb:
Is there interest in a PR to support multiple lines when a variable is delimited by quotes?

arcseldon:
Any update on this?

This issue was mysteriously closed - https://github.com/motdotla/dotenv/issues/196
This workaround is definitely not ideal for certs etc 
https://github.com/motdotla/dotenv/blob/v4.0.0/test/.env#L10

Is there a technical impediment preventing its introduction?

george-norris-salesforce:
Wish there was a better solution for this. The \n workaround is not great/

markbirbeck:
Another use case is SQL queries passed in via environment variables. It would be nice to be able to lay them out neatly, too.

jorgeorpinel:
Hi. I this is definitely needed to match Ruby's dotenv standard (https://github.com/bkeepers/dotenv#multi-line-values).

And for me, the `\n` workaround doesn't even work, `\n` is being sent as part of the value... I'm using the latest version, 7.0.0

PR #486, opened 2020-08-06
feat: add option to allow line breaks in values

### LINKED ISSUE #458: Feature: Support multi-line secrets
There have been multiple issues opened about the lack of newline support for values in the past, some going back a few years. #215 took a stab at this specifically for certificates in env files. And yes, in _most_ cases simply providing a path to a cert as a value does the trick. But not in all situations.

The rather large environment I'm working in involves packaging up an application into a single bundle and shipping that off to AWS via serverless. Now, we could massage things to allow us to also ship this private key to the Lambda function, but then legal would go bananas and tell us that we're not in compliance and there'd be a whole political poopstorm internally. And no one wants that. What it boils down to is that we cannot ship a cert to the app's running destination.

Here's a sample key:value for reference:

```
PRIVATE_KEY="-----BEGIN RSA PRIVATE KEY-----
gobblygoop
-----END RSA PRIVATE KEY-----"
```

Note that this is not specific to certs, but to all multiline values:

```
MULTILINE='hello
goodby
allo'
```

For both, only the first line of the multiline value will be read, even though it's valid to wrap multiline values in quotes. They appear in `process.env` as:

```
PRIVATE_KEY: '"-----BEGIN RSA PRIVATE KEY-----',
MULTILINE: "'hello",
```

We could do what many have resorted to, and that's replacing every newline with `'\n'` in the cert in values in question. But holy crap does that get stupid long and it's extremely prone to errors.

This issue hits home especially hard since (on macs at least) if you place these values into a file and source the file, the system reads those values correctly. I haven't found a compelling argument against dotenv supporting the same.

What's holding dotenv back from being able to support this, when *nix flavors seem to be able to from the shell?

andreialecu:
Closes #458

This adds an option `{ multiline: "line-breaks" }` that allows values to contain new lines. See https://github.com/bkeepers/dotenv#multi-line-values for an additional explanation.

The documentation was updated

Also, in the process I discovered that stubbing `parse` was not working properly in tests, so I changed the exports a bit to fix this and accommodate a new test for this option.


lib/main.js
```diff
@@ -89,11 +115,14 @@ function config (options /*: ?DotenvConfigOptions */) /*: DotenvConfigOutput */
...
     }
+    if (options.multiline != null) {
+      multiline = options.multiline
+    }
   }
 
   try {
     // specifying an encoding returns a string instead of a buffer
-    const parsed = parse(fs.readFileSync(dotenvPath, { encoding }), { debug })
+    const parsed = DotEnvModule.parse(fs.readFileSync(dotenvPath, { encoding }), { debug, multiline })
```
andreialecu, line 125: See https://stackoverflow.com/questions/46124087/sinon-stub-spy-on-local-functions-in-unit-testing for an explanation of why this was necessary

ivan-californias:
Any chance of this getting reviewed & merged in the near future?

This also would had solved #196, #391 and #28 which were closed without a real solution

Asuza:
@motdotla what do you think of this change? It seems like a great addition to a very useful tool.

jcalfee:
The heroku cli (and perhaps the Open Source dockku) supports multi line variables.  The deployment environment requires environment variables for everything because it does not have persistent storage so I ended up as a work-around reading from a any multi-line variable which is a hack (inconsistant with the environment).  Shell's support multi-line too.

multi-line-values looks very easy to implement, it looks like you don't have to worry about escaping:
```bash
PRIVATE_KEY="-----BEGIN RSA PRIVATE KEY-----\nHkVN9...\n-----END DSA PRIVATE KEY-----\n"
```

andreialecu:
If anyone is interested in a different solution for this:

Even though I made this PR, I ended up just using base64 encoded variables. 

If you use VSCode, an extension like https://marketplace.visualstudio.com/items?itemName=adamhartford.vscode-base64 makes it very easy to encode/decode strings in your `.env` file.

You can then use a package like https://www.npmjs.com/package/env-var to manage how you consume the variables in your app:

Your `.env` file:
```
SOME_CERTIFICATE=somebase64encodedstring
```

Code:

```ts
import env from 'env-var';

require('dotenv').config();

const certificate = env
      .get("SOME_CERTIFICATE")
      .required()
      .convertFromBase64()
      .asString()
```

motdotla:
This is a great addition @andreialecu. Thank you. We might make some small modifications but otherwise, this will be something we release relatively soon.

Dunky13:
This is awesome! It would really help me out! Do you also need to update the `.d.ts` file? Or is that done automagically?

wuntusk:
Would really love to see this in.

JohnRachid:
Can we please get this resolved and merged?

motdotla:
I should be getting this merged and released today. Thank you for your incredible patience. 💛

motdotla:
Thank you for this @andreialecu. It's been merged. I will sit on it for a day or two - testing it incrementally by hand. When comfortable then I will release it. Code is very complete good but sometimes edge cases sneak in.

Issue #436, opened 2019-10-29
Feature Request: Allow to override already set envs in the process.env


ThaDaVos:
I saw this issue: https://github.com/motdotla/dotenv/issues/115

But I think this is really something that's needed -> yes I saw the fragment in the FAQ -> but I also thought DotENV automatically did this. I'm currently running into a problem with NuxtJs: We're using `dotenv` in the `nuxt.config.js` to set config options based on ENV variables but today I learned that NuxtJs automatically sets the NODE_ENV variable if it's not set ( https://github.com/nuxt/nuxt.js/blob/dev/packages/cli/src/setup.js#L9-L11 ) - this is causing issues because this is done before `dotenv` loads the .env file, which causes our NODE_ENV to be `production` or `development` based on the scenario in the config file, if an option as below was possible that would really help us:

```javacript
require('dotenv').config({ override: true })

// or specific

require('dotenv').config({ override: [
    'NODE_ENV',
    'TEST_ENV'
] })
```

This way, you could manually enable the overriding of all or certain env variables using the .env file -> this would help a lot of people I think because you **do not** always have control over the environment

heythisispaul:
Does the solution added in [this commit](https://github.com/motdotla/dotenv/pull/155/commits/ae0f81c653a0e71df918a1d79961711cb3c6588b) not solve this issue for you?

dvdbot:
Like I said, I saw the fragment in the FAQ (which that commit refers to)

motdotla:
This request comes in all the time. The behavior is surprising.

In the principle of least surprise, we are going to consider it.

TrentaIcedCoffee:
overriding system's envs can be surprising, but a good hint (ex: env PORT is already defined) can also be helpful.

Gohelraj:
The feature of overriding a specific environment variable becomes necessary to resolve issues occurring due to environment variable name conflict like the one describes here https://github.com/docker-library/redis/issues/53

Dimrok:
What about prefixing vars you want to override with `DOTENV_OVERRIDE_`.

It's a pretty standard way of doing things, like Vue.js uses the `VUE_APP_` prefix to send variables to built applications, gitlab uses `K8S_SECRET_` to send variables to the deployed application, etc.

We could add:

`DOTENV_OVERRIDE_` to override a variable, if present. 
`DOTENV_DUMP_` to dump a variable, if present.

wheresrhys:
[...]

PR #487, opened 2020-08-13
Allow env variables to be overwritten

dstankovd:


wirwolf:
waiting for this feature

lib/env-options.js
```diff
@@ -15,4 +15,8 @@ if (process.env.DOTENV_CONFIG_DEBUG != null) {
   options.debug = process.env.DOTENV_CONFIG_DEBUG
 }
 
+if (process.env.DONENV_CONFIG_OVERWRITE != null) {
```
JonForest, line 18: Does this have a typo here? Shouldn't that be `DOTENV_CONFIG_OVERWRITE`?

carvalholeo: @dstankovd do you have any news?

dstankovd: I'm not sure what do you mean. It should check if there is DONENV_CONFIG_OVERWRITE set as env variable on the host machine and use it's value if present. It's covered in test-env-options.js  and works as expected as far as I can tell.

carvalholeo: I think it's about the typo error (like a grammar or orthographic mistake) in line 18.

It's typed **DONENV**_CONFIG_OVERWRITE, but shouldn't be **DOTENV**_CONFIG_OVERWRITE, with a T?

mirsella: i don't think it's a typo since it's DONENV everywhere, not just here. but i don't know why a N and not a T. maybe a typo at first and after it's the editor autocompletion ?

motdotla:
thanks for the code. we might soon fork this off or borrow from your code. the feature is coming soon.

moving this conversation here: #436

Issue #494, opened 2020-10-26
Feature request: Command substitution


andresespinosapc:
Could you add command substitution, like dotenv for Ruby? https://github.com/bkeepers/dotenv#command-substitution

motdotla:
for now please use https://github.com/motdotla/dotenv-expand for this

PR #495, opened 2020-11-09
Added: feature request: Command substitution #494

bijaydas:


motdotla:
There are additional modules that will expand dotenv variables. Please take a look at those. Thank you for the effort on the PR though!

PR #498, opened 2020-11-24
feat: genericize parse type

gf3:
* allow user to provide optional stricter custom type


motdotla:
We are going to skip on this feature at the moment @gf3 since at the lowest level (the .env file) everything is considered a string.

PR #503, opened 2021-01-16
Allow for `import "dotenv/config";` in Node.js

franciscop:
Currently the officially recommended method for auto-importing works with Babel-like configuration that automatically resolves `dotenv/config` to `dotenv/config.js`. However, Node.js doesn't do this resolution, and instead you have to manually specify the extension like this:

```js
// Babel OK, doesn't work on Node.js
import "dotenv/config";

// Works on Node.js (and Babel etc)
// See https://github.com/motdotla/dotenv/issues/89#issuecomment-587753552
import "dotenv/config.js";
```

A way of fixing this is by defining the export mapping in `package.json` as [explained in Node.js documentation](https://nodejs.org/api/packages.html#packages_package_entry_points):

```js
  "exports": {
    ".": "./lib/main.js",
    "./config": "./config.js"
  },
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,6 +3,10 @@
   "version": "8.2.0",
   "description": "Loads environment variables from .env file",
   "main": "lib/main.js",
+  "exports": {
+    ".": "./lib/main.js",
+    "./config": "./config.js"
+  },
   "types": "types",
   "scripts": {
     "flow": "flow",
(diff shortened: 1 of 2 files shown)


RA80533:
This is a much-needed feature for those of us using more recent releases of Node.js.

@franciscop Are the failing tests related to the change made in package.json?

franciscop:
@RA80533 It seems the current `master` is failing for me locally as well. I just hopefully fixed the tests :)

motdotla:
thank you for this

franciscop:
Thanks for accepting it!

PR #510, opened 2021-04-10
Adding a trim

anooppoommen:
While using the lib I've noticed that the last empty line is also considered for key value decoding
adding a trim just to avoid this.


motdotla:
thank you for this. going to leave this as is for now though until the change can be more carefully considered. for now, removing ending newlines from you .env file.

PR #511, opened 2021-04-21
Feat: Scoped Env Vars

alexandermccormick:
Currently, when a `.env` file is loaded the values are always assigned to `process.env`. This feature adds a `scoped` boolean option to skip the value assignments to `process.env`, and instead only return them in the `parsed` object.

My personal use case for this is wanting to keep the global scope of my project as clean as possible. In my project, I have multiple conenctions to databases and other services that can change randomly, and I didn't want to have unused or unnecessary secrets floating around globally. By keeping them in a local variable, we can let the gc get rid of them when we are done with them.

**All tests should pass on your machine.**

### DIFF
--- a/lib/cli-options.js
+++ b/lib/cli-options.js
@@ -1,6 +1,6 @@
 /* @flow */
 
-const re = /^dotenv_config_(encoding|path|debug)=(.+)$/
+const re = /^dotenv_config_(encoding|path|debug|scoped)=(.+)$/
 
 module.exports = function optionMatcher (args /*: Array<string> */) {
   return args.reduce(function (acc, cur) {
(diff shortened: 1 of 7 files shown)


alexandermccormick:
I didn't mean to throw this at you, so i appologize about that. I was originally going to just write a function for this, but thought others might find this feature useful as well.

alexandermccormick:
Side note: the changes to `tests/test-config-cli.js` look like they were already made in another PR (#511)

motdotla:
this is a creative approach @alexandermccormick and cleanly implemented. That said, we try to keep the features for dotenv to a minimum.

What you can do to still serve your use case is use the lower level `dotenv.parse` method.

https://github.com/motdotla/dotenv#parse

Let me know if that does not work well for you though. More details along those lines will help us better understand if we should add the feature.

PR #513, opened 2021-04-25
Fix small typo in JSDoc comment

ImRodry:
This fixes a typo in the description of the config function where the `@link` had an extra | which is not needed.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -47,7 +47,7 @@ export interface DotenvConfigOutput {
 }
 
 /**
- * Loads `.env` file contents into {@link https://nodejs.org/api/process.html#process_process_env | `process.env`}.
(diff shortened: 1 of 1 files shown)


types/index.d.ts
```diff
@@ -47,7 +47,7 @@ export interface DotenvConfigOutput {
 }
 
 /**
- * Loads `.env` file contents into {@link https://nodejs.org/api/process.html#process_process_env | `process.env`}.
```
motdotla, line 50: @ImRodry can you elaborate why the pipe `|` here causes issues with the documentation?

ImRodry: I see you already merged this so you probably got the answer to this question but the issue was that the | was being put into the JSDoc comment and wasn’t just a separator in the `@link` tag.

PR #516, opened 2021-05-05
Drop node 8 support

motdotla:


jcblw (approved):
Looks good, thanks for the cleanup!

MichaelDeBoey:
This should have been a breaking change

motdotla:
@MichaelDeBoey you are right. I will bump to 9.0.0 shortly to rectify as best as possible. We'll still leave 8.3.0 onward as not supporting 8 despite the mistake - so as not to yank those.

PR #525, opened 2021-05-05
Add package.json to exports

MrSnoozles:
This is a fix for the error: `Package subpath './package.json' is not defined by "exports".`


motdotla:
:+1:

Issue #527, opened 2021-05-06
Feature: Support to parse path along with home directory ~


brelian:
Currently, we don't support parsing an env path with a home director character `~`. e.g.

![Snipaste_2021-05-06_20-02-45](https://user-images.githubusercontent.com/23089063/117295223-285e3980-aea6-11eb-994f-8c750eb366b1.png)

motdotla:
An alternative and immediate solution to this is to do:

```
require('dotenv').config({ path: path.join(os.homedir(), '/.env') })
```

We'll discuss your PR separately as possibly being a useful convenience but for the purpose of this issue the above code example should solves things for your immediate needs.

PR #528, opened 2021-05-06
feat: support to parse home directory

brelian:
## Why we need this PR
This PR is committed to supporting parsing an env path with a home directory prefix since we haven't implemented it yet.

Snipaste_2021-05-06_20-02-45

## Related Issues
- #527

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -23,6 +23,7 @@ type DotenvConfigOutput = {
 
 const fs = require('fs')
 const path = require('path')
+const os = require('os')
 
 function log (message /*: string */) {
   console.log(`[dotenv][DEBUG] ${message}`)
@@ -73,6 +74,10 @@ function parse (src /*: string | Buffer */, options /*: ?DotenvParseOptions */) 
   return obj
 }
 
+function resolveHome(envPath) {
+  return envPath[0] === '~' ? path.join(os.homedir(), envPath.slice(1)) : envPath
+}
+
 // Populates process.env from .env file
 function config (options /*: ?DotenvConfigOptions */) /*: DotenvConfigOutput */ {
   let dotenvPath = path.resolve(process.cwd(), '.env')
@@ -81,7 +86,7 @@ function config (options /*: ?DotenvConfigOptions */) /*: DotenvConfigOutput */ 
 
   if (options) {
     if (options.path != null) {
-      dotenvPath = options.path
+      dotenvPath = resolveHome(options.path)
     }
     if (options.encoding != null) {
       encoding = options.encoding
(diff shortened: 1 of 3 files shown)


motdotla:
This is cool @brelian. Can you tell me a little more about your use case?

For example, instead of using `~` why not do this in your config path:

```
require('dotenv').config({ path: path.join(os.homedir(), '/.env') })
```

brelian:
> This is cool @brelian. Can you tell me a little more about your use case?
> 
> For example, instead of using `~` why not do this in your config path:
> 
> ```
> require('dotenv').config({ path: path.join(os.homedir(), '/.env') })
> ```

Hi @motdotla, thanks for your detailed comment in which case is a good solution. However, sometimes, I make an environment my env path, e.g.
```js
DOTENV_CONFIG_PATH=~/.env node -r dotenv/config my_script.js
``` 

Especially, I need to load those environment variables that locate in the home directory before starting my app since I don't want to put the env file along with my repo. In this way, I could prevent clean it up carelessly.

Thanks,
Pylon

motdotla:
Makes sense. That is a good use case.

What OS are you running on? I run `DOTENV_CONFIG_PATH=~/.env node -r dotenv/config my_script.js` on a mac machine and it finds the `.env` successfully.

brelian:
> Makes sense. That is a good use case.
> 
> What OS are you running on? I run `DOTENV_CONFIG_PATH=~/.env node -r dotenv/config my_script.js` on a mac machine and it finds the `.env` successfully.

I'm working on macOS as well. But a lit bit different is that I run it with WebStorm debugging mode instead of shell. 
image

motdotla:
ok, that is a good use case. I will merge this soon-ish - probably sometime mid-next week if not sooner.

brelian:
> ok, that is a good use case. I will merge this soon-ish - probably sometime mid-next week if not sooner.

That sounds great. Please let me know in case any changes.

bdcarr:
Hello friends! Will this also work if I use it like this?

```javascript
// esm test script
process.env.DOTENV_CONFIG_PATH = `~/.env/brycesflorist.env`
import 'dotenv/config'

// cjs test script
require('dotenv').config({ path: `~/.env/brycesflorist.env` })
```

My use case is mocha test scripts that can be run in multiple test accounts. I use dotenv to pull the URL and access token from a local .env file, so people can pull the script from github and use their own test account to run it manually. Right now we use relative paths from `__dirname` but it's a bit of a hassle doing the `../../../` thing, so I'm stoked to see this PR.

motdotla:
@bdcarr yes, this PR should solve that scenario for you as well.

motdotla:
Released in 10.0.0 https://github.com/motdotla/dotenv/compare/v9.0.2...v10.0.0

PR #539, opened 2021-05-26
fix: dotenv/config typings

max-hk:
This pull request fix typings error thrown by TypeScript when we load `dotenv/config` with dynamic import.

Without this PR, TypeScript will raise an error and fail to compile the following code.

```
await import('dotenv/config');

export {}
```

```
index.ts:1:14 - error TS7016: Could not find a declaration file for module 'dotenv/config'. '/path/to/project/node_modules/dotenv/config.js' implicitly has an 'any' type.
  Try `npm i --save-dev @types/dotenv` if it exists or add a new declaration (.d.ts) file containing `declare module 'dotenv/config';`
```


danny-abc:
@motdotla can you take a look on this?

motdotla:
Yes, will do so this weekend. Thank you for this.

motdotla:
@max-hk can you elaborate on using `import('dotenv/config')` in typescript rather than..

```js
import dotenv from "dotenv"
dotenv.config()

console.log(process.env)
```

config.js is mainly to be used for preloading.

motdotla:
Added to v13.0.0 release: https://github.com/motdotla/dotenv/blob/master/CHANGELOG.md#1300-2022-01-16

Issue #548, opened 2021-06-30
Dotenv running on debug mode for any non-undefined debug option


n8eloy:
Hey all! We have noticed that passing any defined values in the `debug` option in `config` function will turn on debug mode.

e.g. all those will activate debug logs:
```ts
dotenv.config({ debug: true })
dotenv.config({ debug: false })
dotenv.config({ debug: 'foo' })
dotenv.config({ debug: 0 })
dotenv.config({ debug: {} })
```

The [documentation](https://github.com/motdotla/dotenv/blob/master/README.md#debug) mentions that `debug` defaults as false, but I didn't expect that passing false explicitly wouldn't make it work the same way.

What I'm exactly trying to accomplish is programmatically activate debug mode for a certain environment variable:

```ts
dotenv.config({ debug: process.env.NODE_ENV === 'development' })
```

The only way I've made it work was by passing an undefined or null value:

```ts
dotenv.config({ debug: process.env.NODE_ENV === 'development' ? true : undefined })
```

I do think this happens because `options.debug` is checked against `null`, but not against falsy values. Is this expected behavior?

https://github.com/motdotla/dotenv/blob/27dfd3f034ce00b1daa72effbd91dd7788aced48/lib/main.js#L83-L97

I'm using Dotenv version `8.6.0`

nifr:
This issue is still present in `10.0.0`. 

@motdotla  - Can I help somehow to get #550 merged and a new release tagged?

PR #550, opened 2021-06-30
fix: dotenv/main debug option not work 

189:
fix the issue [548](https://github.com/motdotla/dotenv/issues/548) 
All those will active debug mode:
```
  dotenv.config({ debug: true })
  dotenv.config({ debug: 1 })
  dotenv.config({ debug: 'some string' })
```

And belows will  not:

```
  dotenv.config({ debug: false })
  dotenv.config({ debug: null })
  dotenv.config({ })
  dotenv.config({ debug: 0 })
```

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -91,7 +91,7 @@ function config (options /*: ?DotenvConfigOptions */) /*: DotenvConfigOutput */ 
     if (options.encoding != null) {
       encoding = options.encoding
     }
-    debug = Boolean(options.debug);
+    debug = Boolean(options.debug)
   }
 
   try {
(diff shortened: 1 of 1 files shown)


bigmeow:
I think this code change is better than your`s
```js
  let dotenvPath = path.resolve(process.cwd(), '.env')
  let encoding /*: string */ = 'utf8'

  if (options) {
    if (options.path != null) {
      dotenvPath = resolveHome(options.path)
    }
    if (options.encoding != null) {
      encoding = options.encoding
    }
  }
```

lib/main.js
```diff
@@ -91,9 +91,7 @@ function config (options /*: ?DotenvConfigOptions */) /*: DotenvConfigOutput */
     if (options.encoding != null) {
       encoding = options.encoding
     }
-    if (options.debug != null) {
-      debug = true
-    }
+    debug = Boolean(options.debug)
```
Jolg42, line 94:
```suggestion
```
Suggested by @bigmeow

motdotla:
Thank you for this. I will additionally modify to @bigmeow's recommendation after merging.

Issue #554, opened 2021-07-18
fs module dependency error Next JS


Ajay-056:
### From NPM JS for "fs" package
This package name is not currently in use, but was formerly occupied by another package. To avoid malicious use, npm is hanging on to the package name, but loosely, and we'll probably give it to you if you want it.

### ERROR OCCURING
error - ./node_modules/dotenv/lib/main.js:24:0
Module not found: Can't resolve 'fs'

### QUICK FIX
Using  "fs-extra" package from NPM JS instead of "fs" package.

PR #555, opened 2021-07-18
fs module dependency fix

Ajay-056:


Ajay-056:
Refer the below issue for more insights

https://github.com/motdotla/dotenv/issues/554

PR #558, opened 2021-07-22
feat: drop Node 10 support

### LINKED ISSUE #557: Drop Node 10 support
Because it's [EOL](https://nodejs.org/en/about/releases).

MichaelDeBoey:
BREAKING CHANGE: Node@>=12 is required

---

Closes #557

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,9 +7,9 @@ os:
   - windows
 
 node_js:
-  - "10"
-  - "12"
-  - "14"
+  - 12
+  - 14
+  - 16
 
 cache: npm
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
+## [Unreleased](https://github.com/motdotla/dotenv/compare/v10.0.0...master)
+
+### Changed
+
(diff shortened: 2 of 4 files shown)


pan93412 (approved):
LGTM. But would we use the features that Node 10 didn't bring to us?

jimmywarting:
> LGTM. But would we use the features that Node 10 didn't bring to us?

how about switching to esm instead?

UltiRequiem:
> > LGTM. But would we use the features that Node 10 didn't bring to us?
> 
> how about switching to esm instead?

This would be the best

jimmywarting (approved):

UltiRequiem (approved):

motdotla:
Thank you for this @MichaelDeBoey

motdotla:
> > > LGTM. But would we use the features that Node 10 didn't bring to us?
> > 
> > 
> > how about switching to esm instead?
> 
> This would be the best

Considering this. 

Would anyone like to chime in on potential downsides?

Fotiman:
> > > > LGTM. But would we use the features that Node 10 didn't bring to us?
> > > 
> > > 
> > > how about switching to esm instead?
> > 
> > 
> > This would be the best
> 
> Considering this.
> 
> Would anyone like to chime in on potential downsides?

Downside: there are probably many existing applications that still rely on commonjs, and therefore would no longer be able to upgrade this package if it switches to esm.

motdotla:
Yes, we have considered ESM. We will move to it at sometime but this is a very heavily relied on library and as a result we are going to maintain the status quo at this time.

Issue #552, opened 2021-07-14
new feature proposal: default value if var is not defined


stefanocudini:
how can I make sure that if an environment variable is not defined(inside .env file) I can have a default value?
example:
assuming that the PORT variable is not specified in my .env file
but I want to have a default value anyway:
```javascript
require('dotenv').config({
  defaults: {
    PORT: 8080
  }
})
```

paramsiddharth:
Looks good! If approved, I would like to work on the issue.

tawfiek:
Looking cool 👌🏻

motdotla:
@stefanocudini, @paramsiddharth, @tawfiek can you elaborate on the use case for this?

Why not just make sure the ENV is set in the `.env` file? Why have to go to 2 places to figure out where a value is being injected?

tawfiek:
@motdotla in my case we ignore the .env file from remote repos, adding these default values in the code itself might make the developers start the service in there development machine without having the hassle to share the .env file with each other, or making out form .env.example or something like this.

And making the .env file specific to override the default values in the remote environments, specifically for these tiny service that doesn't contain these much secrets.

it might makes the onboarding a little bit easier, just clone, serve and you good to go.

motdotla:
> without having the hassle to share the .env file with each other, or making out form .env.example

@tawfiek would [Dotenv Sync](https://sync.dotenv.org/) solve that for you?

tawfiek:
@motdotla yea, looks promising actually. 🧐
Interesting to get more details about how it's working.

> > without having the hassle to share the .env file with each other, or making out form .env.example
> 
> @tawfiek would [Dotenv Sync](https://sync.dotenv.org/) solve that for you?

motdotla:
@tawfiek These docs can get you started quickly: https://docs.dotenv.org/#guides

And then we also have quite a few video guides here: https://vimeo.com/dotenv

motdotla:
@tawfiek I think your team will like it. It makes it very easy to securely sync you `.env` files across your team's many machines. So far everyone using it so far is really enjoying it. They say it is like using `git` - one of the first things they now reach for with their team.

PR #563, opened 2021-09-19
feat: Add support for default values

smyles96:
Added support for default values when a variable is undefined as proposed within issue #552.

Changes:
- dotenv parse function changed to support a new `options.defaults` property. Specified defaults will be added to a final parsed object only if respective key was not encountered within `.env` file.
- Updated TAP tests to test new default value functionality and adjust to the new number of log calls made within function.
- README's options subsection updated to show usage of the **defaults** option.

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,6 +124,18 @@ You may specify the encoding of your file containing environment variables.
 require('dotenv').config({ encoding: 'latin1' })
 ```
 
+#### Defaults
+
+Default: `{}`
+
+You may specify default environment variables to inject when not found in your `.env` file.
+If a variable key is found in both your `.env` file and your default values, the value
+in your `.env` file will override the default value.
+
+```js
+require('dotenv').config({ defaults: { PORT: 3000 } })
+```
+
 #### Debug
 
 Default: `false`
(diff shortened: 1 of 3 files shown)


motdotla:
This is pretty solid put together @smyles96, thank you... but, we should get some clarification first on the use case. 

I added a comment here: https://github.com/motdotla/dotenv/issues/552#issuecomment-1014141674

motdotla:
I've thought this over and my biggest concern is it adds confusion to the developer. Now the environment variables are hidden in 2 places - in the code (as defaults) and in a `.env` file. Developers have built a lot of muscle memory over the years in regards to `.env` files. I don't think we should make them go a second place to find environment variables.

That said, I understand the problem this is solving for - insecurely passing around `.env` files or using a `.env.example` file. That has always been cumbersome, but now you can use https://github.com/dotenv-org/cli. It syncs your files across machines and team members - solving the problem very elegantly and securely.

motdotla:
So in short:

1. Use https://github.com/dotenv-org/cli
2. Closing this PR (still open to further discussion even while closed)

Issue #545, opened 2021-06-10
comment after var assignment deemed invalid


Morriz:
Hi, I think this is valid but invalid does not like it:

SOME_VAR=true # bla

It sources fine, it's just that invalid does not trim comments before parsing I guess.

meduzen:
Seems to be a duplicate of #524, am I right?

Morriz:
indeed...closing!

Issue #524, opened 2021-05-05
Feature: Support inline comments


motdotla:
Adding comments to a .env file would be very helpful and improve developer UX.

* https://github.com/motdotla/dotenv/issues/156
* https://github.com/motdotla/dotenv/issues/484
* https://github.com/bkeepers/dotenv#comments

should require a major version bump

brelian:
Hi @motdotla, how is it going on this feature? Does anyone take it yet? If nobody, I would like to try to address it.

Issue #484, opened 2020-08-05
Remove comments while parsing/Support inline comments


JaneJeon:
In VSCode, when you install the `dotenv` extension, it greys out sections following a hashtag, e.g.:

```
SOME_VAR=HELLO # explanation of SOME_VAR, or why it's assigned the value HELLO
               ^--------------------this part is greyed out------------------^
```

However, `dotenv` _the package_ does not respect that. In fact, I often got burned by the fact that dotenv just straight up includes the comments in the actual value for the environment variable, which is super unintuitive.

I'd love it if I could have inline comments without fear of screwing up the environment variables.

Love the work you're doing,
thanks!

fastfedora:
I'll like to second this request. I just got bit by this expectation as well. 

Issue #156 addressed this, but was closed saying that dotenv is expected to be just like a .ini file. Except that it's in the format of Unix shell variables, and most shells support inline comments in their configuration files. So the expectation is that this is supported. 

Nor is the issue immediately obvious, since `dotenv` does not throw an error, but merely includes the comment as part of the variable value, so it can take a while to track down the issue as a `.env` file issue.

dotnetCarpenter:
I got the same issue. After searching for comment syntax, I found https://github.com/vlucas/phpdotenv#comments which is a PHP implementation that allows comments after the key-value. I think it is important to keep parity between `.env` implementations so that the same `.env` file can be used across different implementations. E.i. [Ruby dotenv](https://github.com/bkeepers/dotenv), [PHP dotenv](https://github.com/vlucas/phpdotenv) and [Nodejs Dotenv](https://github.com/motdotla/dotenv/).

Original Ruby Dotenv comment syntax: https://github.com/bkeepers/dotenv#comments

motdotla:
I agree that dotenv should support comments. It follows the principle of least surprise and makes for a nice developer experience. I am closing this issue and instead tracking here: https://github.com/motdotla/dotenv/issues/524

PR #568, opened 2021-10-15
Support inline comments

jonathan-reisdorf:
fixes https://github.com/motdotla/dotenv/issues/524
fixes https://github.com/motdotla/dotenv/issues/545
fixes https://github.com/motdotla/dotenv/issues/156
fixes https://github.com/motdotla/dotenv/issues/484


motdotla:
@PodaruDragos no, nothing blocking. We are just prioritizing some cleanup and tightening up of the project before this. This will be a major new upgrade.

motdotla:
@jonathan-reisdorf thank you for this and for your patience. Apologies for the delay. If you end up signing up joining [Dotenv Sync](https://sync.dotenv.org), let me know and I will make sure they get you 6 months free for your organization.

renandecarlo:
This broke using # in the variable itself. Such as: 
MSG=Hello, you've got a new order (#%order_n%)

How should I use it now?

jonathan-reisdorf:
@renandecarlo Yes, unquoted `#` characters are now the beginning of a comment. In order to use the character in the value, you now have to put it in quotes, e.g.
```dotenv
MSG="Hello, you've got a new order (#%order_n%)"
```

jonathan-reisdorf:
(We could also take into account if there is a space before the `#` character. E.g. the GNU coreutils' `env` does that. Then again, I've seen other frameworks/tools behaving exactly as dotenv is now. Let me know @motdotla if you want the `#` only to mark the beginning of a comment if it's preceded by a space character or if it's the first character, then I'm happy to create a PR for this)

motdotla:
Thanks @jonathan-reisdorf. Let's leave it as is. That has the least surprise.

Sorry about this breaking your line @renandecarlo, but, as @jonathan-reisdorf mentioned, just put double quotes around it and you will be good to go.

renandecarlo:
I had to stick with version 13, otherwise it would break my program for everyone already using it without quotes. But that's ok.
I think most comments would have had a space before # though

motdotla:
Ah darn. As a developer myself I never enjoy getting stuck like that. We'll consider an optional flag where you can switch to an older mode. 

When you say all your devs would have to switch, can you elaborate on that? You mean they would each have to update their own .env file locally?

renandecarlo:
Yes. The example env file shipped with some default values which contained `#`, like the example above, and those could be changed or not. Everyone would then have to add the quotes to it, otherwise it would break.

Devtronyx:
> Thanks @jonathan-reisdorf. Let's leave it as is. That has the least surprise.
> 
> Sorry about this breaking your line @renandecarlo, but, as @jonathan-reisdorf mentioned, just put double quotes around it and you will be good to go.

It is still treated as a comment even inside double quotes. I'm on version 14.2.0. And the value represents a password so it's not much I can do about it

motdotla:
@jonathan-reisdorf can you double check that, please, if you get a chance before I do?

jonathan-reisdorf:
> > Thanks @jonathan-reisdorf. Let's leave it as is. That has the least surprise.
> > Sorry about this breaking your line @renandecarlo, but, as @jonathan-reisdorf mentioned, just put double quotes around it and you will be good to go.
> 
> It is still treated as a comment even inside double quotes. I'm on version 14.2.0. And the value represents a password so it's not much I can do about it

See https://github.com/motdotla/dotenv/issues/601#issuecomment-1019563032
Typically the wrapping quotes would be outside of the value

jonathan-reisdorf:
> When you say all your devs would have to switch, can you elaborate on that? You mean they would each have to update their own .env file locally?

@motdotla if @renandecarlo uses dotenv the same way as we are in our organization, there's a `.env.example` file that's committed to git containing the default values, and developers copy that file to their own local `.env` which is gitignored.
That way, devs can make local changes to the values without creating unstaged changes potentially containing secrets which they might accidentally commit. The disadvantage to this is that there's usually no migration/syncing mechanism between the two files.

motdotla:
> The disadvantage to this is that there's usually no migration/syncing mechanism between the two files.

Indeed. We've been building out [Dotenv Sync](https://sync.dotenv.org/) to help solve this problem. @renandecarlo you can give that a try. Then you could make the updates to the `.env` file (adding quotes), and then all your team members would just run `dotenv-cli pull` to instantly update their local `.env` file.

PR #571, opened 2021-11-14
explicitly require process

UltiRequiem:
https://nodejs.org/api/process.html#process

> While it is available as a global, it is recommended to explicitly access it via require or import.


motdotla:
Can you explain why this is needed for the repo beyond it is just recommended? Until then, marking closed.

PR #578, opened 2021-12-25
Fix TypeScript types nodenext support

xiaoxiangmoe:
See config example at https://devblogs.microsoft.com/typescript/announcing-typescript-4-5-beta/#packagejson-exports-imports-and-self-referencing

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,10 @@
   "description": "Loads environment variables from .env file",
   "main": "lib/main.js",
   "exports": {
-    ".": "./lib/main.js",
+    ".": {
+      "default": "./lib/main.js",
+      "types": "./types/index.d.ts"
+    },
     "./config": "./config.js",
     "./config.js": "./config.js",
     "./package.json": "./package.json"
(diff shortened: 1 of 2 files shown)


motdotla:
this is great. thank you. will merge and be part of the next release soon.

package.json
```diff
@@ -4,7 +4,10 @@
   "description": "Loads environment variables from .env file",
   "main": "lib/main.js",
   "exports": {
-    ".": "./lib/main.js",
+    ".": {
+      "default": "./lib/main.js",
```
motdotla, line 8: Before I merge this:

can you point me to the definition of using `default` rather than specifying both `import` and `require`? I cannot find the documentation on this.

motdotla: I also wonder if moving our type definition file under lib would solve a lot of these edge cases - from the past and going forward. That was the historical default and most editors I believe are set up to look there.

xiaoxiangmoe: doc of `default`, `import` and `require` is here.
https://nodejs.org/api/packages.html#conditional-exports

I think it will be better if we moving our type definition file under `lib`. Should I create another PR to do this?

motdotla: I agree. Can you just roll it into this PR, please? Thank you for your help!

xiaoxiangmoe: @motdotla see https://github.com/motdotla/dotenv/pull/585

Issue #601, opened 2022-01-22
# inside double quotes still treated as inline comment


Devtronyx:
v14.2.0
`SOME_KEY={"field1":"some value", "field2": "stops here#this is discarded"}`
When above line is parsed generates the following output
`SOME_KEY: {"field1":"some value", "field2": "stops here`
which in turns gives a SyntaxError: Unexpected end of JSON input

motdotla:
Thank you for raising this. We are looking into it. Anyone else this is happening to, please +1.

jonathan-reisdorf:
@Devtronyx interesting. In this case, even though it's a json this is read as a string. You'd have to quote the entire value in order to properly parse this (in your case since you already use the double quotes I suggest to use the single ones as wrapper):

```
SOME_KEY='{"field1":"some value", "field2": "stops here#this is discarded"}'
```

kirillgroshkov:
Broke us too.

I don't know though what's more "correct" behavior - to require a space before `#` to treat it as comment or not.

jonathan-reisdorf:
Quick collection of patterns I noticed in other tools since yesterday:

vscode: syntax highlighting suggests: unquoted `#` is treated as comment, with or without space
<img width="166" alt="image" src="https://user-images.githubusercontent.com/8958624/150789682-e81fb85e-6f39-4d9e-b214-3ac49bde86f9.png">

Github:
```env
FOO=bar#baz
FOO=bar #baz
FOO="bar#baz"
FOO="bar #baz"
FOO='bar#baz'
FOO='bar #baz'
```

python-dotenv started out where we are but with version 0.10.3 they introduced a check for space character

unix (GNU coreutils) env I've already shared here: https://github.com/motdotla/dotenv/pull/568#issuecomment-1018402106

mrliptontea:
Just ran into this. It's very common for generated secrets to include `#` so something like:

```shell
VAR_WITH_SPECIAL_CHARS=space slash/other@#1!=
```

But unfortunately:

```diff
--- Expected
+++ Received

 {
-  VAR_WITH_SPECIAL_CHARS: "space slash/other@#1!="
+  VAR_WITH_SPECIAL_CHARS: "space slash/other@"
 }
```

motdotla:
@mrliptontea yes, it is very common. We are working on the long term solution for this. Until then, you can do:

1. Downgrade to `13.x`
[...]

PR #603, opened 2022-01-25
Interpret # as start of comment only if preceded by whitespace

jonathan-reisdorf:
It looks like https://github.com/motdotla/dotenv/pull/568 introduced a breaking change for more people/organizations than I anticipated. This PR changes this so that inline comments have to start with a whitespace character before the `#`, making it much less likely for people upgrading from dotenv < version 14 to break things. Technically, this is again a breaking change, though, for people already using version 14 and using `#` as an inline comment without a preceding whitespace character 😅 

fixes https://github.com/motdotla/dotenv/issues/601

(feel free to close, e.g. in case you have something else in mind like a feature flag for inline comments)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ All notable changes to this project will be documented in this file. See [standa
 
 ## [Unreleased](https://github.com/motdotla/dotenv/compare/v14.3.0...master)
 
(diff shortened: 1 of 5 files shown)


motdotla:
Thanks @jonathan-reisdorf.

And yes, technically a breaking change, but I'm going to release this as a patch since the feature is so new and effecting people immediately.

Issue #605, opened 2022-01-25
Calling dotenv.config() results in a TypeError


bankslenus:
Since upgrading from dotenv 14.2.0 to 14.3.0 I get the following error when calling `dotenv.config()`

CODE (foo.js):
```
...
import * as dotenv from "dotenv";
dotenv.config(); // line 20
...
```

ERROR:
```
$ yarn build
$ tslint -c tslint.json -p tsconfig.json --fix
$ tsc
$ node --es-module-specifier-resolution=node .
file:///..../foo.js:20
dotenv.config();
       ^

TypeError: dotenv.config is not a function
    at file:///..../foo.js:20:8
    at ModuleJob.run (node:internal/modules/esm/module_job:185:25)
    at async Promise.all (index 0)
    at async ESMLoader.import (node:internal/modules/esm/loader:281:24)
    at async loadESM (node:internal/process/esm_loader:88:5)
    at async handleMainPromise (node:internal/modules/run_main:65:12)
```

Downgrading to 14.2.0 solves the error.

JeromeFitz:
Downgrading to `14.2.0` is an option. Thanks for identifying and suggesting @bankslenus

`14.3.0` exports as `default` currently: [`main.js`](https://github.com/motdotla/dotenv/blob/a1bf6c1371c8c9c5ab51f9ecb08ebfae46007910/lib/main.js#L140).

```sh
[Module: null prototype] {
  default: { config: [Function: config], parse: [Function: parse] }
}
```

For now you can do:

```js
import dotenv from "dotenv";
```

_Aside:_

In one of my cases I was dynamically importing (only loading on local development not CI) and needed to change to this:

```js
...
  const dotenv = await import('dotenv')
  dotenv.default.config({ path: './.env' })
...
```

Though this would work too: 

```js
...
  const {
    default: { config },
  } = await import('dotenv')
  config({ path: './.env' })
...
```

---

The code change looks to come in from this [commit](https://github.com/motdotla/dotenv/commit/8875300c181974ee24d742876fedf4c4e6e94251#):

**Before**:

```js
module.exports.config = config
module.exports.parse = parse
```

**After**:
```js
const DotenvModule = {
   config,
   parse
 }

 module.exports = DotenvModule
 ```

Will see if we can double-up the `module.exports` as a `patch` for the interim to see if going to `14.3.1` is an option for backward compatibility, or if a different direction is preferred.

motdotla:
[...]

PR #606, opened 2022-01-25
patch for 14.3.0, add back explicit module.exports #605

JeromeFitz:
To preserve backwards compatibility with `14.3.0` this re-introduces the exports in place prior, while keeping the new default.

Details of issue are in #605

**Please note:**
I can see this PR not being needed. The changes introduced in `14.3.0` could be considered the responsibility of the implementer, not `dotenv`. 

In which case this could be considered a `README` update (separate PR), not a patch or breaking change.

If this is accepted, I would see these lines of code not existing in the next breaking version. 😆

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -137,4 +137,6 @@ const DotenvModule = {
   parse
 }
 
+module.exports.config = DotenvModule.config
+module.exports.parse = DotenvModule.parse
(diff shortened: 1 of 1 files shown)


motdotla:
Thank you for the contribution @JeromeFitz 💛

PR #609, opened 2022-01-28
docs: improve explanation about environment in React

Josh-Cena:
As a maintainer of [Docusaurus](https://github.com/facebook/docusaurus), I was asked recently about how dotenv works with React. Then I realized a new section was added to the documentation recently (I haven't used dotenv myself for a year now), which unfortunately is very... misleading. I hope this elaboration makes it easier to understand for new React developers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -311,7 +311,11 @@ require('dotenv').config({ override: true })
 
 ### How come my environment variables are not showing up for React?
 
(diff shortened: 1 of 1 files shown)


motdotla:
Thank you for this added clarification @Josh-Cena.

PR #610, opened 2022-01-29
Update env-options.js

Muffle10:
It would be a bit simpler if we changed it to this instead of all of that unneeded logic


motdotla:
Hi @Muffle10, thank you for the PR. We are going to leave things as they are though. There is an edge case where `false` could be set on an ENV - in which case we want the value of `false` to pass through to options.

PR #615, opened 2022-02-02
Add support for backticks

motdotla:


s100:
Was it necessary to bump the major version number for this? Normally, adding a new feature isn't a breaking change.

T0biii:
@s100 i think it is a breaking change see also: "If you had values containing the backtick character, please quote those values with either single or double quotes."
https://github.com/motdotla/dotenv/blob/master/CHANGELOG.md

motdotla:
@s100 it is a good point you raise. The difficulty with the nature of this repo (and we err on the side of safety) is that a regex change on the parser can break edge cases for users of the older version. 

For example, in this case there could be users out there with a value like this:

`` KEY=ab`cd ``

mo:
@motdotla what are the security implications of backtick support?

motdotla:
@mo great question, Mo. Also great short handle!

There is no execution of variables (as there is in node template variables), so there are no additional security implications here.

The backticks simply act as a parsing wrapper - like single or double quotes.

PR #628, opened 2022-02-18
Update the broken reference link to require module

kunalpanchal:
The existing link `https://nodejs.org/api/cli.html#cli_r_require_module` in the Readme.md is broken. Fixing it to be https://nodejs.org/api/cli.html#-r---require-module instead.


motdotla:
Thank you for this 😀

Issue #633, opened 2022-02-23
Inline comments may cause impossible to set values


Toilal:
I have a quite secure token generated with shiny characters, including `"`, `'` `` and `#`.

```
6U+c'UDH"l`ZFDD5%/|'t{Ojt.5hzu+#wUBH#:9w*l_I2z{^m/7h-U&!qcLlXZ_
```

The new inlined comment feature is great, but it makes this kind of token impossible to declare inside `.env` file, even using any of the supported quoting character (single quote, double quote, backtick ...)

See unit test in the related PR.

Toilal:
Another unit tests seems weird too, and it seems to be related.

https://github.com/motdotla/dotenv/blob/0e004970f08da84c6e381d7715182da2b23af311/tests/.env#L32

https://github.com/motdotla/dotenv/blob/0e004970f08da84c6e381d7715182da2b23af311/tests/test-parse.js#L66

Expectation seems to be wrong here as `#space` should be kept in this case as previous character is not a space.

motdotla:
Oh my. So sorry about this. Will look into it promptly.

psoleckimoj:
Any progress on this as it seems to still happen?

`TEST=abc#def` gets read in as `abc`

The issue is that escaping doesn't work properly either, so I escape it `abc\#def` and it gets read in as `abc\#def`? For some reason it's always leaving the escape character in the string?

We're having to create a list of characters not to use when generating secrets purely because of this issue when trying to dev locally. Doesn't happen when pulling secrets in k8s as it correctly removes the escape characters.

motdotla:
@psoleckimoj - use quotes and you won't have to escape:

```
TEST="abc#def"
```

MathRobin:
I have the same problem, and double-quoted don't work. Same for things with %, like `%26` (for example) into string value, it converts to hexa code so, it's replaced with `&`.

motdotla:
at some point in time we fixed this since i am unable to duplicate - even when using with dotenv-expand.

here's a working example here:

https://github.com/dotenv-org/examples/blob/master/usage/dotenv-nodejs-inline-comments-expand/.env


<img width="995" alt="Screenshot 2024-02-05 at 3 23 05 PM" src="https://github.com/motdotla/dotenv/assets/3848/f7a6b219-01f2-4707-8a84-74afc6927e95">

PR #634, opened 2022-02-23
Add test with impossible token

Toilal:
See #633


rmclaughlin-nelnet:
This is a great start. It would be even better if all special characters were included to make sure all edge cases are handled.

PR #636, opened 2022-02-27
Add support for base64 encoded environment config

saptarshi1996:
dotenv has support has Buffer but not for base64. So I added support for it. 

This will make the environment variables more secure. 

Eg, There is a config 

SERVER=localhost\nPASSWORD=password\nDB=tests\n.
Then it gets encoded to U0VSVkVSPWxvY2FsaG9zdFxuUEFTU1dPUkQ9cGFzc3dvcmRcbkRCPXRlc3RzXG4=

On decoding  and parsing it we get the required key value pairs. 

 /^([A-Za-z0-9+/]{4})*([A-Za-z0-9+/]{3}=|[A-Za-z0-9+/]{2}==)?$/g
This regex will test if the config is in base64 form. 

Then we decode it using src = Buffer.from(src, 'base64').

I have added test cases for multiple cases.

Please let me know, if any other changes are required for this update

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -8,6 +8,15 @@ const LINE = /(?:^|^)\s*(?:export\s+)?([\w.-]+)(?:\s*=\s*?|:\s+?)(\s*'(?:\\'|[^'
 function parse (src) {
   const obj = {}
 
+  // Check if src is base64 encoded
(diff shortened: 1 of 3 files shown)


saptarshi1996:
@motdotla

PR #637, opened 2022-03-01
Add support for base64 encoded string

saptarshi1996:
dotenv has support has Buffer but not for base64. So I added support for it.

This will make the environment variables more secure.

Eg, There is a config

SERVER=localhost\nPASSWORD=password\nDB=tests\n.
Then it gets encoded to U0VSVkVSPWxvY2FsaG9zdFxuUEFTU1dPUkQ9cGFzc3dvcmRcbkRCPXRlc3RzXG4=

On decoding and parsing it we get the required key value pairs.

/^([A-Za-z0-9+/]{4})*([A-Za-z0-9+/]{3}=|[A-Za-z0-9+/]{2}==)?$/g
This regex will test if the config is in base64 form.

Then we decode it using src = Buffer.from(src, 'base64').

I have added test cases for multiple cases.

Please let me know, if any other changes are required for this update

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -8,6 +8,15 @@ const LINE = /(?:^|^)\s*(?:export\s+)?([\w.-]+)(?:\s*=\s*?|:\s+?)(\s*'(?:\\'|[^'
 function parse (src) {
   const obj = {}
 
+  // Check if src is base64 encoded
+  const base64Test = /^([A-Za-z0-9+/]{4})*([A-Za-z0-9+/]{3}=|[A-Za-z0-9+/]{2}==)?$/g
+  const isBase64Encoded = base64Test.test(src)
+
+  // If true, then decode base64
+  if (isBase64Encoded) {
+    src = Buffer.from(src, 'base64')
+  }
+
   // Convert buffer to string
   let lines = src.toString()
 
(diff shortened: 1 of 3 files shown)


saptarshi1996:
@motdotla please let me know if this is can be added

motdotla:
Hi @saptarshi1996 this is interesting but can you elaborate on the use case some more?

Currently, I don't see a real need for this. Maybe others can also chime in.

saptarshi1996:
Hello. @motdotla. 

So, This comes from a requirement of mine. 
Suppose we could base64 encode the entire the data in the .env file for added security. If someone gets access to your .env file all they see is a base64 encoded text.

motdotla:
> or when you're pair programming with an outside user and wish to hide your password or secret keys this will be useful

So I like your thinking @saptarshi1996. It's innovative and it would be nice as a convenience.

Take a look at what we are doing with https://github.com/dotenv-org/cli - it's early stage but it is going to eventually allow fully encrypted .env files. 

Until then, we won't merge this PR, but thank you for the creative thinking. Consider creating your own module here - maybe call it dotenv-md5 and make it wrap this library (similar to how dotenv-expand works)

motdotla:
So all that said, I am going to close this PR without merging.

saptarshi1996:
@motdotla thanks, would like to contribute at https://github.com/dotenv-org/cli,

PR #644, opened 2022-03-14
Feature/esm export

ckcr4lyf:
A bug in either node/cjs lexer causes ES6 imports of CJS to fail in some shared environments (steps to reproduce are including in the issue I made here -> https://github.com/nodejs/node/issues/42222)

Although this should ultimately be "solved" in node itself, I was wondering if this ESM export is a potential step dotenv is willing to take.

I'd opened a discussion (Related to this issue - https://github.com/motdotla/dotenv/discussions/640), but didn't get any feedback, so thought an MR might be the next step.

This does fix the memory issue for me, when I install dotenv from my fork.

### DIFF
--- a/lib/main.mjs
+++ b/lib/main.mjs
@@ -1,0 +1,109 @@
+import fs from 'fs';
+import path from 'path';
+import os from 'os';
+
+const LINE = /(?:^|^)\s*(?:export\s+)?([\w.-]+)(?:\s*=\s*?|:\s+?)(\s*'(?:\\'|[^'])*'|\s*"(?:\\"|[^"])*"|\s*`(?:\\`|[^`])*`|[^#\r\n]+)?\s*(?:#.*)?(?:$|$)/mg
+
+// Parser src into an Object
+function parse (src) {
+  const obj = {}
+
+  // Convert buffer to string
+  let lines = src.toString()
+
+  // Convert line breaks to same format
+  lines = lines.replace(/\r\n?/mg, '\n')
+
+  let match
+  while ((match = LINE.exec(lines)) != null) {
+    const key = match[1]
+
+    // Default undefined or null to empty string
+    let value = (match[2] || '')
+
+    // Remove whitespace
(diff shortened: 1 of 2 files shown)


motdotla:
Thanks for the PR @ckcr4lyf. One concern we have with this is maintaining 2 files that essentially implement the same thing here.

Thoughts on that?

ckcr4lyf:
I agree, this MR specifically probably isn't the best solution since it leads to duplication with manual copy paste, and maintaining changes as such. 

In some of my research, I came across [tsup](https://github.com/egoist/tsup) which can bundle the same typescript source into both ESM and CJS. 

However considering the size & popularity of dotenv, I was guessing it's been a conscious decision to not introduce compilation from typescript (though it is already a dev dependency). If you are open to the idea, I would be happy to explore a typescript source based build step which leads to producing both `esm` and `cjs` form a single source file!

motdotla:
Your guess is right. We'd rather not get into a build process. Copy/paste isn't the worst thing.

We'll let this sit right now and will consider it over the next few weeks. 

Anyone else please chime in with your thoughts as well.

ckcr4lyf:
Btw: Got feedback from node team - they say this is "intended behavior" - i.e. the lexer for using CJS in ESM reserves a lot of memory, which means in restricted ulimit environments it will not run - https://github.com/nodejs/node/issues/42222#issuecomment-1067950915

So I think at the moment, for anyone using dotenv in a "pure ESM" project (i.e. excluding dependencies), that needs to be run in shared environments with ulimit restrictions, an ESM export fork is required.

ckcr4lyf:
Update: guybedford over at node team added a fallback parser for importing CJS in an ESM package via https://github.com/nodejs/node/pull/43612 

So on latest node (v18.6.0) I do not have this issue with dotenv. Since it is an official release, I will close this MR for now, though I do think it would be nice to have an ESM export for dotenv in the future.

PR #651, opened 2022-04-07
Parsed file entries can be undefined

### LINKED ISSUE #650: DotenvParseOutput declares that all environment variables are defined
As the code currently exists the type system can't help the user realize that the values coming from the parse might not exist.

Example:
```ts
import FS from "fs";
import * as DotEnv from "dotenv";

const envVals = DotEnv.parse(FS.readFileSync("myvars.env"));

console.log(`MY_ENV_VAR type: ${typeof envVals.MY_ENV_VAR}`);
// Typescript says the type of envVals.MY_ENV_VAR is "string", but what is it really?
```

Save the above into `test.ts`, execute as follows:
```sh
touch myvars.env
$(npm bin)/ts-node-script test.ts
```

My result:
```text
MY_ENV_VAR type: undefined
```

Thus there's a disparity between what actually happens and the type system, causing the developer to not be aware of an undefined value until runtime.

Yes I know I could just load into process.env using the `config()` method and get the signature I'm looking for - but my current use case calls for loading the vars without placing them into process.env hence the use of `parse`.

kf6kjg:
Assuming that the file contains every entry you put in the code is untrue. This change brings the output type into line with NodeJS process.env.

BREAKING CHANGE: Code that makes the assumption that the values are defined will now fail to compile. Validate your inputs.

Fixes #650

### DIFF
--- a/lib/main.d.ts
+++ b/lib/main.d.ts
@@ -1,9 +1,7 @@
 // TypeScript Version: 3.0
 /// <reference types="node" />
 
-export interface DotenvParseOutput {
-  [name: string]: string;
-}
+export type DotenvParseOutput = NodeJS.Dict<string>
 
 /**
  * Parses a string or buffer in the .env file format into an object.
--- a/tests/types/test.ts
+++ b/tests/types/test.ts
@@ -2,7 +2,7 @@ import { config, parse } from "dotenv";
 
 const env = config();
 const dbUrl: string | null =
-  env.error || !env.parsed ? null : env.parsed["BASIC"];
+  env.error || !env.parsed ? null : env.parsed["BASIC"] ?? "";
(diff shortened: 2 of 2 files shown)


kf6kjg:
I marked this as a breaking change due to an abundance of caution. If you wish to have me downgrade to a simple fix instead, please let me know and I can force-push an amendment.

motdotla:
hi @kf6kjg, thank you so much for the insight here and for the PR. I left it up to see if it would gather a lot more response from the community. It hasn't and my gut has been to avoid doing this. 

If you create an npm module that wraps dotenv to do this though, please place here and I will add to dotenv's README. similar to how dotenv-expand adds to dotenv's functionality.

PR #654, opened 2022-05-02
Changed regex groups to named groups

mainarthur:
Changed regex groups to named groups to improve readability of regex and code


motdotla:
Hi @mainarthur, this is improved readability is nice, but I'm very cautious with this repo and going to keep it as is. I'm extra paranoid that these kinds of changes might cause issues on older node versions. We try to minimize that as much as possible unless the PR adds significant value to override that. I hope you understand.

PR #655, opened 2022-05-04
Add a load function for loading .env without applying it

dobesv:
It seemed unexpectedly hard to load and parse the `.env` file from the default path (or `$DOTENV_CONFIG_PATH`).  This seemed like a pretty simple refactor that could come in handy from time to time.

In my case I wanted to apply a whitelist to the results, so I could use this function and filter out the keys that I don't want before applying to `process.env`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,7 +103,6 @@ You can load the `.env` file without updating `process.env` using the `load` fun
 ```javascript
 const dotenv = require('dotenv')
 const config = dotenv.load() // will return an object
-console.log(typeof config, config) // object { BASIC : 'basic' }
 ```
 
 ### Preload
--- a/README.md
+++ b/README.md
@@ -98,6 +98,15 @@ const config = dotenv.parse(buf) // will return an object
 console.log(typeof config, config) // object { BASIC : 'basic' }
 ```
 
+You can load the `.env` file without updating `process.env` using the `load` function:
+
+```javascript
+const dotenv = require('dotenv')
+const buf = Buffer.from('BASIC=basic')
+const config = dotenv.load() // will return an object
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -98,6 +98,15 @@ const config = dotenv.parse(buf) // will return an object
 console.log(typeof config, config) // object { BASIC : 'basic' }
 ```
 
+You can load the `.env` file without updating `process.env` using the `load` function:
+
+```javascript
+const dotenv = require('dotenv')
+const buf = Buffer.from('BASIC=basic')
```
ritschwumm, line 105: looks like `buf` is unused here...

dobesv:
```suggestion
```

README.md
```diff
@@ -98,6 +98,14 @@ const config = dotenv.parse(buf) // will return an object
 console.log(typeof config, config) // object { BASIC : 'basic' }
 ```
 
+You can load the `.env` file without updating `process.env` using the `load` function:
+
+```javascript
+const dotenv = require('dotenv')
+const config = dotenv.load() // will return an object
+console.log(typeof config, config) // object { BASIC : 'basic' }
```
ritschwumm, line 106: the `// object { BASIC : 'basic' }` comment probably has to be removed, too

dobesv, line 106:
```suggestion
```

motdotla:
@dobesv Appreciate this PR, but I think it's easy enough for someone to do:

```js
dotenv.parse(fs.readFileSync('.env'))
```

I'm going to close this, but do please elaborate more on why making this a first-class method would be worthwhile. Might reopen.

dobesv:
> why making this a first-class method would be worthwhile

Hmm I can't remember any more, seems fine to close this.

PR #667, opened 2022-08-19
fix broken esm description

mifi:


kev009:
Yeah this ends up being more natural, it seems to run early in import.  You can use something like `const myVar = process.env.MYVAR` in top level scope of a file.  The current documented method of an early call does not allow this.

motdotla:
Thank you @mifi

PR #668, opened 2022-08-20
better way to use config()

danilluk1:
I think there is a better way to import config function. It solves a bug, that happens when some dependencies will have null in .env file, because dotenv module is doesn't loaded
```javascript
import * as dotenv from 'dotenv' // see https://github.com/motdotla/dotenv#how-do-i-use-dotenv-with-import
dotenv.config()
import "somemodule" //if in this module we needs environment we may have an error, that happens because to this moment dotenv isn't loaded because of es6 module loading peculiarities
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,8 +43,7 @@ console.log(process.env) // remove this after you've confirmed it working
 .. or using ES6?
 
 ```javascript
-import * as dotenv from 'dotenv' // see https://github.com/motdotla/dotenv#how-do-i-use-dotenv-with-import
-dotenv.config()
+import "dotenv/config" from "dotenv" //better way to config in es6
 import express from 'express'
 ```
 


kev009:
Yeah this ends up being more natural, it seems to run early in import.  You can use something like `const myVar = process.env.MYVAR` in top level scope of a file.  The current documented method of an early call does not allow this.

kev009:
https://github.com/motdotla/dotenv/pull/667 seems to have the same idea with a bit more doc coverage.

klark6130:
I found that this is working https://github.com/dotenv-org/examples/blob/master/dotenv-es6-import-pitfall/valid.mjs

motdotla:
Thank you for this PR. We merged another PR that made a similar update to the README. Closing this one.

PR #675, opened 2022-09-16
refactor: change old while match to #String.matchAll

thejimbow:
I think it's better to use KISS here with #String.matchAll and iterate over matched key values.


motdotla:
Thank you for digging into the internals. We're going to leave as is though.

PR #676, opened 2022-09-19
Add Stricter Encoding type

MRDGH2821:
The motive is to introduce stricter type & offer autocomplete in code editors as this change directly gathers possible encoding choices from Node.js itself

### DIFF
--- a/lib/main.d.ts
+++ b/lib/main.d.ts
@@ -35,7 +35,7 @@ export interface DotenvConfigOptions {
    *
    * example: `require('dotenv').config({ encoding: 'latin1' })`
    */
-  encoding?: string;
+  encoding?: BufferEncoding;
 
   /**
    * Default: `false`


motdotla:
I'm hesitant to go stricter. dotenv has always had a spirit of flexibility and looseness.

Maybe you can elaborate more for me why this would be useful. I'm not a heavy typescript user. I use it when I can. I prefer duck typed languages.

MRDGH2821:
Over here you can see the suggestions poping up:
image

Those are here thanks to:
image

It is same for `readFileSync()` as well

But it is not here in:
image

Due to: 
image

Ultimately `config()` uses `readFileSync()` 
image

And thus I suggested this PR.

If the user does put incorrect encoding, `fs` module will anyways throw the error. So to tell them available correct encodings, stricter types can be introduced which is ultimately used by `fs` module either ways.

Also this type is directly derived from node so there is less maintainance overhead.

motdotla:
Sorry, can you elaborate in more detail? Why is it so bad to have all those suggestions pop up?

I am a dynamic language guy so don't know this typed world as well. What's wrong with string | undefined ?

MRDGH2821:
My point is to have those suggestions popup while giving encoding option in `config({encoding: '' })`

Although there is no harm in having `string | undefined`, but when `fs.readFileSync()` itself is providing the options, why shouldn't `config()` as well?

Those suggestions come up as a result of providing `BufferEncoding` 
Thus the PR.

To check it out how those suggestion popup come, open my PR or the main repo in vscode.

Import `fs.readFileSync()` in a file and while providing `encoding` option, press `ctrl` + `space` (or do it after putting 2 single quotes & moving the cursor between the two) 

Ultimately my PR was trying to enhance developer experience, but at the same time I respect your decision to reject this PR

PR #686, opened 2022-10-20
Update package.json

SumitJain01:


cheng-alvin:
@SumitJain01  You should provide more information regarding what you are doing.

cheng-alvin (changes requested):
Hey! As seen [here](https://www.npmjs.com/package/standard-markdown), the latest version is 7.1.0. Maybe revert this back?

cheng-alvin:
@motdotla We can close this now

PR #689, opened 2022-10-26
More characters for env keys

GijsvandenHoven:
I noticed some of my env variables were not being loaded at all. After some prodding, It turned out a regex was the culprit, allowing only alphanumeric, underscore, dot and hyphen for variable names.

This PR extends the Regex to allow any kind of bracket (round, square, angle, curly).

I also updated the ReadMe section on how the parser works to specify precisely what kind of characters are allowed in variable names.

### DIFF
--- a/README.md
+++ b/README.md
@@ -330,6 +330,7 @@ The parsing engine currently supports the following rules:
 
 - `BASIC=basic` becomes `{BASIC: 'basic'}`
 - empty lines are skipped
+- variable names must consist of alphanumeric characters and `_.-{}[]()<>`
 - lines beginning with `#` are treated as comments
 - `#` marks the beginning of a comment (unless when the value is wrapped in quotes)
 - empty values become empty strings (`EMPTY=` becomes `{EMPTY: ''}`)
--- a/lib/main.js
+++ b/lib/main.js
@@ -5,7 +5,7 @@ const packageJson = require('../package.json')
 
 const version = packageJson.version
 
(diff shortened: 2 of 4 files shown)


motdotla:
Thank you for this @GijsvandenHoven. It's interesting. I haven't come across this format in my own experience.

Can you elaborate on it a bit? Why is this format used at times?

GijsvandenHoven:
@motdotla Thank you for looking at my PR.

For myself, I ended up needing this format when my .env file specified a global variable to apply, but I also needed to specify a number of exemptions overriding this 'default' with something else.

I tried to do this with square brackets for syntactic sugar reasons, leading me to find out this wasn't possible. Something along the lines of `OVERRIDE_FOO[12]=bar`. (I worked around it using `.` instead).

So I would call the general use case that falls under, some kind of dictionary registering where specific values deviate from a default.

This obviously better lends itself to something like a JSON object, but this wasn't really an option for our circumstances.

Not being able to do this stumped me at first, because the only specification on rules for environment variable names I could find was [this](https://pubs.opengroup.org/onlinepubs/000095399/basedefs/xbd_chap08.html), stating most characters on the keyboard other than `=` are fine.

motdotla:
I'm hesitant to merge this because I'm afraid of complexities downstream - what if a platform like Vercel or someone else does not support KEYs like that when setting environment variables for production?

Were you also using this format on your production servers?

GijsvandenHoven:
That's a valid concern. I do not know anything about platforms like Vercel. 

Instead of using this fork on production servers, I ended up just using `.` to separate variable parts of the key when parsing the environment variables. Having keys with dynamic sections that are parsed is a strange scenario anyway, it should be JSONs job.  I was just in a very particular scenario where this was needed, and the brackets were a syntactic sugar choice.

If it's at the cost of _anything_ working properly, this sugar isn't worth it.

motdotla:
i'm going to close this. thank you though for the PR and the interesting approach.

PR #695, opened 2022-11-12
update README

klark6130:


motdotla:
Thank you for this. We merged another PR that had the same fix.

PR #700, opened 2022-12-06
Add env() function to get key value 

baselrabia:
using env function is more easier and more readable for getting the value from enviroment vaiable 

before 
```js
require("dotenv").config();

const db_pass = process.env.DB_PASS
console.log(db_pass);

```

after 

```js
const env = require("dotenv").env;

const db_pass = env("DB_PASS")
console.log(db_pass);

```

### DIFF
--- a/lib/main.d.ts
+++ b/lib/main.d.ts
@@ -71,3 +71,14 @@ export interface DotenvConfigOutput {
  *
  */
 export function config(options?: DotenvConfigOptions): DotenvConfigOutput;
+
+/**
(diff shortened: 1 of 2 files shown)


motdotla:
Thank you for the PR. Always appreciate new ideas.

That said, we are not going to do this.

You can do effectively the same thing with this:

```js
const result = require('dotenv').config()
const env = result.parsed
console.log(env)
console.log(env["HELLO"])
```

```
# .env
HELLO="World"
```

https://github.com/dotenv-org/examples/blob/master/dotenv-assignment/index.js

PR #713, opened 2023-02-02
feat: basic escaping support

### LINKED ISSUE #711: Supporting Escape Sequences
Hello @motdotla  and thank you for creating dotenv nodejs library 👍 

## Feature Request

Could dotenv nodejs library support escape sequences (e.g `\" \t`) ?

At the minimum can it be aligned with the Ruby dotenv library?
- [version 2.8.0](https://github.com/bkeepers/dotenv/releases/tag/v2.8.0) added https://github.com/bkeepers/dotenv/pull/421

**Would a PR which aligns with the Ruby library be welcome?**

## Context
I've noticed some implementations of dotenv support character escapes and some do not.
Also there does not seem to be an official specification anywhere for dotenv.
This is causing issues when we need to use the same `.env` files in multiple runtimes.

Is the Ruby library the **original** implementation and thus also used as the **pseudo specification**?

### [Ruby Library](https://github.com/bkeepers/dotenv)
- Supports escaping anything other than `$`
- Only interpret values of `\n` and `\r` and only in double quoted strings.
- [Relevant Source Code](https://github.com/bkeepers/dotenv/blob/0974810e909076cc34d5579ac119c9422e5f64b1/lib/dotenv/parser.rb#L74-L94)

### [Elixir Library escaping specs](https://hexdocs.pm/dotenvy/dotenv-file-format.html#escape-sequences)
- Seems to allow expanding of more special sequences versus the Ruby library, e.g: `\t` and `\f`
- Does not seem to distinguish (at the spec level at least) between single or double quotes escape values.

### [Python Library](https://github.com/theskumar/python-dotenv#file-format)
- ```
     Allowed escape sequences:

    in single-quoted values: \\, \'
    in double-quoted values: \\, \', \", \a, \b, \f, \n, \r, \t, \v
  ```

bd82:
This PR implements basic escaping support as implemented in the Ruby Gem
- https://github.com/bkeepers/dotenv/blob/0974810e909076cc34d5579ac119c9422e5f64b1/lib/dotenv/parser.rb#L70-L98

I have attempted to align the code style / logic components with the Ruby Gem.
Based on the assumption that this is the "source of truth" (pseudo spec) for this library.

I personally think other implementations of dotenv have more complete (better?) support for escaping and
**expanding** special characters ([Example Spec](https://hexdocs.pm/dotenvy/dotenv-file-format.html#escape-sequences). 

But I wanted to start with the minimal viable scope and avoid misalignment with the "source of truth" (?)

resolves  #711

### DIFF
--- a/tests/test-parse-multiline.js
+++ b/tests/test-parse-multiline.js
@@ -23,9 +23,9 @@ t.equal(parsed.DOUBLE_QUOTES_SPACED, '    double quotes    ', 'respects surround
 
 t.equal(parsed.EXPAND_NEWLINES, 'expand\nnew\nlines', 'expands newlines but only if double quoted')
 
-t.equal(parsed.DONT_EXPAND_UNQUOTED, 'dontexpand\\nnewlines', 'expands newlines but only if double quoted')
+t.equal(parsed.DONT_EXPAND_UNQUOTED, 'dontexpandnnewlines', 'expands newlines but only if double quoted')
 
-t.equal(parsed.DONT_EXPAND_SQUOTED, 'dontexpand\\nnewlines', 'expands newlines but only if double quoted')
+t.equal(parsed.DONT_EXPAND_SQUOTED, 'dontexpandnnewlines', 'expands newlines but only if double quoted')
 
 t.notOk(parsed.COMMENTS, 'ignores commented lines')
 
--- a/lib/main.js
+++ b/lib/main.js
@@ -22,28 +22,41 @@ function parse (src) {
     const key = match[1]
 
     // Default undefined or null to empty string
-    let value = (match[2] || '')
-
-    // Remove whitespace
-    value = value.trim()
-
-    // Check if double quoted
-    const maybeQuote = value[0]
-
-    // Remove surrounding quotes
-    value = value.replace(/^(['"`])([\s\S]*)\1$/mg, '$2')
-
-    // Expand newlines if double quoted
-    if (maybeQuote === '"') {
-      value = value.replace(/\\n/g, '\n')
-      value = value.replace(/\\r/g, '\r')
-    }
-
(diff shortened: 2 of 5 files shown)


tests/test-parse-multiline.js
```diff
@@ -23,9 +23,9 @@ t.equal(parsed.DOUBLE_QUOTES_SPACED, '    double quotes    ', 'respects surround
 
 t.equal(parsed.EXPAND_NEWLINES, 'expand\nnew\nlines', 'expands newlines but only if double quoted')
 
-t.equal(parsed.DONT_EXPAND_UNQUOTED, 'dontexpand\\nnewlines', 'expands newlines but only if double quoted')
```
motdotla, line 26: i have mixed feelings on this behavior. looks like the rubygem has done this since inception, but it doesn't follow the principle of least surprise to me.

tests/test-parse-multiline.js
```diff
@@ -23,9 +23,9 @@ t.equal(parsed.DOUBLE_QUOTES_SPACED, '    double quotes    ', 'respects surround
...
+t.equal(parsed.DONT_EXPAND_UNQUOTED, 'dontexpandnnewlines', 'expands newlines but only if double quoted')
```
motdotla, line 26: i have mixed feelings on this behavior. looks like the rubygem has done this since inception, but it doesn't follow the principle of least surprise to me. the `\n` disappearing would frustrate me.

motdotla: this behavior goes back to `0.0.3`. https://github.com/motdotla/dotenv/commit/599144a3d2e22f6539a9429d409f1723769736b6

i'm open to arguments for doing this, but it will definitely be a potentially significant breaking change if we do

Issue #709, opened 2023-01-15
dotenv/safe as replacement for dotenv-safe


karlhorky:
Hi @motdotla, first of all, thanks so much for creating and maintaining `dotenv`! 🙌 It's a great tool, and our students get a lot of value from it.

I wanted to test the waters with how open you would be to absorbing the functionality of [`dotenv-safe`](https://www.npmjs.com/package/dotenv-safe) (to verify that all keys from `.env.example` have matching keys in the `.env` file) under a separate, opt-in import path, eg:

```ts
require('dotenv/safe').config()
// or
import * as dotenv from 'dotenv/safe'
```

`dotenv-safe` is no longer maintained (last publish 3 years ago) and [depends on a very old version of `dotenv`](https://github.com/rolodato/dotenv-safe/issues/128).

It seems like most of the functionality of `dotenv-safe` is [here](https://github.com/rolodato/dotenv-safe/blob/1ab395c2045bc8a2737da1a683cb6d56605a08f4/index.js#L21-L48).

It seems like a good number of `dotenv` users are using this `.env.example` paradigm in the wild:

- https://github.com/motdotla/dotenv/issues/437

This would enable those users to also verify that their `.env` file configures all of the environment variables from `.env.example` file that is committed to version control. This can prevent things like an unconfigured environment variable causing a production runtime error. I know there are alternatives to using `.env.example` including eg. sharing encrypted secrets, but the `.env.example` pattern is simple, relies more on existing tools and workflows, and is widely used.

chav-aniket:
+1 to this absolutely would love this feature in dotenv

tenlisboa:
@karlhorky I forked the [`dotenv-safe`](https://github.com/rolodato/dotenv-safest) and created a separated package with the updated dependencies, so we can maintain this package, until it's not a part of dotenv core module.

Check it out: [dotenv-safest](https://www.npmjs.com/package/dotenv-safest)

motdotla:
Awesome. Thank you @tenlisboa!

karlhorky:
@motdotla are you open to bringing this into `dotenv` core too at some point?

motdotla:
Yes definitely but not prioritizing just yet. There are many upvotes for it as well.

tenlisboa:
@motdotla If you don't mind, I can incorporate that feature, and you can only review, validate and merge. What do you say?

motdotla:
[...]

PR #714, opened 2023-02-08
feat: dotenv-safe as a feature of dotenv in dotenv/safe

ParadoxInfinite:
Hi @motdotla, thank you for dotenv! I have been a fan and a user for a while now!

When I saw issue https://github.com/motdotla/dotenv/issues/709, it made sense to make this a feature of dotenv and as the code for dotenv-safe was already there, I took the opportunity to integrate it into dotenv.

I haven't worked with TAP tests before, so I am having a bit of tough time getting the tests right, however, this wouldn't affect any existing features, so I will be updating this soon. Could you take a look at the functionality of the code and if it makes sense? Most of the code is taken from https://github.com/rolodato/dotenv-safe, so huge props to @rolodato as well for dotenv-safe!

Let me know if this is not in the idea of the package and if this won't be implemented, I would then extend dotenv or dotenv-safe to make a custom version available with the latest dotenv package.

Thanks! 🍻 

---

resolves https://github.com/motdotla/dotenv/issues/709


ParadoxInfinite:
I've cleaned up a bit and added a couple of tests. I might be missing a few test cases, but I assume this would suffice for now.
I have an issue with one of the lines in which I am not sure how to test, so if someone could help me out with it, that'll be great!

motdotla:
hi @ParadoxInfinite, thank you so much for the PR but i'm going to close this. i've reached out to @rolodato to request taking over maintenance of dotenv-safe. would prefer that than add this to dotenv.

PR #719, opened 2023-02-21
chore: reorder the types field in package.json

YuzeTT:
Should be the first in the object as required by TypeScript. This is more conducive to TypeScript work.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,8 +6,8 @@
   "types": "lib/main.d.ts",
   "exports": {
     ".": {
-      "require": "./lib/main.js",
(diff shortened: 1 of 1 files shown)


cheng-alvin (approved):
Looks good!

motdotla:
> This is more conducive to TypeScript work.

How? Can you elaborate? Not sure why TypesScript would have an opinion on this either way.

motdotla:
I will merge since it does not matter to me either way, but I would like to know why this would matter to TypeScript.

PR #720, opened 2023-02-26
feat: Accept URL as path option

ukstv:
Since Node.js v7.6.0 `fs.readFileSync` accepts `URL` as a path argument.

This PR amends type definition to reflect that: `path` option is declared as `string | URL`.


motdotla:
https://github.com/dotenv-org/examples/pull/9

ukstv:
Thank you very much!!

Issue #690, opened 2022-10-28
Decomposition proposal


krutoo:
Hi, i want use `dotenv` for parse and merge objects manually, as it is done under the hood of this library.

I want to extract mechanism of modifying `process.env` by parsing result.

Can we figure out how to decompose `config()` function into for example `config()` and `merge()` or `apply()`.

In this case `config()` will be looks like:

```js
function config (options) {
  // ...
  try {
    const parsed = parse(fs.readFileSync(dotenvPath, { encoding }));

    apply({ 
      target: process.env,
      source: parsed,
      options: { debug, override },
    });

    return { parsed }
  } catch (err) {
  // ...
}
```

The `apply()` function may be looks like:

```js
function apply ({ target, source, options }) {
  const { override, debug, log } = options;

  for (const key of Object.keys(source)) {
    if (!Object.prototype.hasOwnProperty.call(target, key)) {
      target[key] = source[key];
    } else {
      if (override === true) {
        target[key] = source[key];
        debug && log(`"${key}" is already defined and WAS overwritten`);
      } else {
        debug && log(`"${key}" is already defined and was NOT overwritten`);
      }
    }
  }
}
```

motdotla:
This is a good suggestion. Putting on todos.

deepam-kapur:
Can I work on this @motdotla ? And can you assign me this?

motdotla:
Let's give @krutoo the chance to do so. @krutoo, I agree with this proposal. It's a great one. Want to knock it out and I'll review and maybe make some small modifications.

deepam-kapur:
ok sure. let me know If I could help too! @motdotla &  @krutoo

motdotla:
@deepam-kapur if we don't here from @krutoo in 10 days then go ahead and start on the PR and we'll credit him in the changelog.

krutoo:
i am here

krutoo:
@deepam-kapur you can do it if you want

motdotla:
Ok, great. @deepam-kapur go ahead and work on a PR for that

deepam-kapur:
sure thanks!

deepam-kapur:
@motdotla I've created a PR please review and let me know if any change is required.

deepam-kapur:
@motdotla have you got chance to review this?

motdotla:
@deepam-kapur reviewed and merged. it will be released as part of 16.1.0 at the end of the month.

motdotla:
[...]

PR #733, opened 2023-05-11
[FEATURE] apply function added where we can pass target, source and o…

deepam-kapur:
…ptions to parse and fill the data. Closes - #690

#690 Apply function added

args -
target - process.env or any other object
source - object parsed with parse function
options - same as config (no change)

### DIFF
--- a/lib/main.d.ts
+++ b/lib/main.d.ts
@@ -61,6 +61,34 @@ export interface DotenvConfigOutput {
   parsed?: DotenvParseOutput;
 }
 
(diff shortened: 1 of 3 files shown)


krutoo:
will the "read and parse dotenv" function be implemented? Or is it already in the package?

motdotla:
Nice PR @deepam-kapur.

I'm going to merge. I might make some small changes and possibly change the name `apply` to something else, but this is great work. The community will love this. Thank you!

PR #736, opened 2023-05-15
Add config usage example code when using ES6 import to README.md

HwangTaehyun:
Add config usage example code when using ES6 import


motdotla:
Thank you for this, but I think the docs below explain this well enough but in the standard js format. ES6 users can modify to ES6 format as they wish.

PR #751, opened 2023-06-13
fix missing URL import in types

Jakousa:
**What:**

Adds missing import to types.

**Why:**

The types are missing import which results in
```
node_modules/dotenv/lib/main.d.ts:29:19 - error TS2304: Cannot find name 'URL'.

29   path?: string | URL;
```

See https://github.com/DefinitelyTyped/DefinitelyTyped/issues/19799.

### DIFF
--- a/lib/main.d.ts
+++ b/lib/main.d.ts
@@ -1,5 +1,6 @@
 // TypeScript Version: 3.0
 /// <reference types="node" />
+import type { URL } from 'node:url';
 
 export interface DotenvParseOutput {
   [name: string]: string;


motdotla:
The types stuff is never ending. Is there a way to do this in the package.json instead of adding code?

TylerHaigh:
Hi all,

Since this change we have observed the following error in our application


```
node_modules/dotenv/lib/main.d.ts:3:26 - error TS2307: Cannot find module 'node:url' or its corresponding type declarations.

3 import type { URL } from 'node:url';
```

Is there something additional we need to add to make this work?

TylerHaigh:
Update: Our project was using `@types/node@13` - bumping this to at either `@types/node@14` or `@16` works
(`@18` didn't work)

motdotla:
Thank you. For anyone else experiencing @TylerHaigh's issue do the same for now.

I might consider reverting this PR. It is getting annoying how type changes have so many side effects for others - causing perfectly fine working software to cause failures.

Abramovick:
I'm also facing this error!

PR #757, opened 2023-07-03
Can get env variables null-safety

tenlisboa:
In the projects I worked on, I had to create an abstract class or a helper function to validate if the environment variable existed before using it for security and time-saving purposes. 

So I implemented a built-in get accessor for environment variables that primarily does three things:
- Get an env value from the processEnv option or process.env object.
- Parse boolean envs by default
- Throws an exception if the environment is not defined (it saves a lot of debugging time instead of looking up your code to see why the integration is not working when you just git pushed the code).

### DIFF
--- a/lib/main.d.ts
+++ b/lib/main.d.ts
@@ -154,3 +154,15 @@ export function populate(processEnv: DotenvPopulateInput, parsed: DotenvPopulate
  *
  */
 export function decrypt(encrypted: string, keyStr: string): string;
+
+/**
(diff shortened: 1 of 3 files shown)


tenlisboa:
@motdotla Do you think it would be useful for dotenv purposes?

tenlisboa:
Closing this PR. No response back

motdotla:
It's a cool idea but I don't think it should be part of core dotenv - at least not at this time.

Create a wrapper/extension lib for it - like dotenv-expand.

PR #761, opened 2023-07-18
Fix: Redundant code removal.

joyson-madtha0:
I was stuck integrating `dotenv` with `vitest` on `nextjs` app. While doing reverse engineering on `dotenv` to figure out what I am doing wrong, I stumbled on to this piece of code. Hope this fix will help a little.

Anyway, I did integrated `dotenv` with `vitest`. This change is not a blocker for me...

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -303,12 +303,4 @@ const DotenvModule = {
   populate
 }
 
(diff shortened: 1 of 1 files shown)


motdotla:
This has some purpose with different import vs require approaches. Going to leave as is. Thank you for the PR though.

PR #780, opened 2023-10-31
Fix weird freezing on import

Lucaszw:
Importing dotenv after registering swc / ts-node seems to freeze indefinitely (I suspect it's related to how its handled w/ es6) . This seems to resolve the issue.


motdotla:
Could you post example code. I haven't been able to replicate any freezing.

PR #792, opened 2023-12-30
Fix return type of populate()

### LINKED ISSUE #790: DotenvPopulateOutput typing is incorrect
The `populate` function has no return statement, so it cannot return an object of the form that `DotenvPopulateOutput` describes.

For reference:
```typescript
export interface DotenvPopulateOutput {
  error?: Error;
}

/**
 * [...]
 * @returns {void}
 */
export function populate(/* ... */): DotenvPopulateOutput;
```

Should be:
```typescript
export function populate(/* ... */): void;
```

phuvinhbmt:
Resolves #790


motdotla:
thank you

Issue #796, opened 2024-01-23
Using an array for config.path does not work


tran-simon:
According to the [readme](https://github.com/motdotla/dotenv?tab=readme-ov-file#path), the `path` option should accept an array

> path
Default: `path.resolve(process.cwd(), '.env')`
By default, config will look for a file called .env in the current working directory. Pass in multiple files as an array, and they will be loaded in order. The first value set for a variable will win.
`require('dotenv').config({ path: ['.env.local', '.env'] })`

It does not seem to be working properly

In this snippet, `process.env.TEST` is undefined when using `path: [".env"]`
```
console.log("Loading with array");
require("dotenv").config({ path: [".env"] });
console.log(process.env.TEST);

console.log("Loading with path");
require("dotenv").config({ path: ".env" });

console.log(process.env.TEST);
```
Output:
> Loading with array
undefined
Loading with path
test

[...]

PR #797, opened 2024-01-23
fix: support array for path option

tran-simon:
Fixes: #796


motdotla:
this gets us halfway there. it selects the first file. merging and then i will add to it to properly support smart loading each file in order.

motdotla:
published as `16.4.1` https://www.npmjs.com/package/dotenv?activeTab=versions. i'll work toward the full fix (where multiple file merge together) as soon as possible and will be released as a minor version.

PR #805, opened 2024-02-07
Process multiple files in options.path, if provided.

thanosd:
I cleaned up the code a bit to first gather a list of files (that exist) in a uniform way.  The logic is as close as to what existed:  

1. If `options.path` has one or more options use those.  
2. If none were provided, or none of them existed, use the default `.env`.

The next step is to parse and combine those values - flattening them.  The sequence would be:

existing env -> options.path[0] (if exists) -> options.path[1] -> etc 

There were two choices here:

1. Move from left to right, and if something exists "first one wins" unless `override` is set.
2. Start from options.path[0] and move right using first one wins (no overrides), then take the combo result and apply it to `existing env` honoring the `override` flag.

I went with (2) as I did not deem the `override` semantics made sense when collapsing multiple environment files - but let me know if you disagree.

Also cleaned up the docs to reflect this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -336,7 +336,9 @@ Specify a custom path if your file containing environment variables is located e
 require('dotenv').config({ path: '/custom/path/to/.env' })
 ```
 
-By default, `config` will look for a file called .env in the current working directory. Pass in multiple files as an array, and they will be loaded in order. The first value set for a variable will win.
+By default, `config` will look for a file called .env in the current working directory.
+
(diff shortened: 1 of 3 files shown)


motdotla:
@thanosd can you switch to

> Move from left to right, and if something exists "first one wins" unless override is set.

that is what `dotenvx` is doing. 

thank you for all your hard work here @thanosd. it's really contributing to making dotenv much better for everyone.

thanosd:
Updated as requested.  Little had to change in the code structure, as we need to hold a copy of the combined parsed object to return it.

motdotla:
thank you. i should merged and released today.

PR #811, opened 2024-02-13
fix: Set NodeJS v14 as the oldest supported engine version

lekspyl:
dotenv v16.4.3 can't be used with Node 12 because the change introduced in the PR https://github.com/motdotla/dotenv/pull/805 contains an optional chaining operator that is supported by NodeJS starting from v14, thus running it with v12 engine version results in the
```
  if (options?.encoding) {
              ^
SyntaxError: Unexpected token '.'
```
exception.

Considering that Node v12 has reached EOL I assume it's better to set v14 as the oldest version constraint.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -21,7 +21,7 @@
         "typescript": "^4.8.4"
       },
       "engines": {
-        "node": ">=12"
+        "node": ">=14"
       },
       "funding": {
         "url": "https://dotenvx.com"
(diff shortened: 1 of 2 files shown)


motdotla:
maybe, but it would be nice to still support older versions. many people still run node 12 in production. @thanosd would you like to take a crack at modify that to support node 12?

motdotla:
tests passed here for 12. https://github.com/motdotla/dotenv/pull/812

and released under `v16.4.4`.

thank you for contributing this @lekspyl. i'm going to close though.

thanosd:
You beat me to it.

PR #814, opened 2024-02-20
do not check if exists

motdotla:


motdotla:
@thanosd just a heads up, i'm releasing this. i should have caught it in your pr but didn't. historical behavior is that if `path` is set it does not attempt to default back to `.env` - even if a file in the listed paths is not found.

PR #815, opened 2024-02-21
README: change intended location of .env file from project root to working directory

stefanneuhaus:
Although this is correctly explained further down in the section on Config options, being not 100% correct in this "Hello World" example can cause unnecessary confusion in the first place.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,7 +83,7 @@ Or installing with yarn? `yarn add dotenv`
 </div>
 </a>
 
-Create a `.env` file in the working directory of your process:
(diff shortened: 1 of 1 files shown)


stefanneuhaus:
I agree, having a simple introductory example is nice. In "my" case an undergraduate of mine spent hours analyzing the issue for her project with a project structure like `backend/app.js`. This might be an unusual structure, but still.
Maybe there is a better phrase that is both simple and less misleading in those cases.

motdotla:
can you change the text to something like:

```
...eate a `.env` file in the root of your project. (if using a monorepo structure like `apps/backend/app.js`, you should put in the root of your backend folder where your app.js process runs)
```

stefanneuhaus:
I updated the text accordingly.

PR #819, opened 2024-03-06
Update package.json

vit100:
Removed "main" key.

The "exports" provides a modern alternative to "main" allowing multiple entry points to be defined, conditional entry resolution support between environments, and **preventing any other entry points besides those defined in "exports"**

If both ["exports"](https://nodejs.org/api/packages.html#exports) and ["main"](https://nodejs.org/api/packages.html#main) are defined, the ["exports"](https://nodejs.org/api/packages.html#exports) field takes precedence over ["main"](https://nodejs.org/api/packages.html#main) in supported versions of Node.js.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,6 @@
   "name": "dotenv",
   "version": "16.4.5",
   "description": "Loads environment variables from .env file",
-  "main": "lib/main.js",
   "types": "lib/main.d.ts",
   "exports": {
     ".": {


motdotla:
thank you for the PR but this will break for thousands (possibly even hundreds of thousands of people) on old versions of node still.

vit100:
should not as package.json clearly says
"engines": {
    "node": ">=12"
  },
  and node >12 supports "exports" key.

motdotla:
indeed. you are right but we don't hold to it strictly. there are many companies still on older versions using this library so we try to avoid modern changes. for faster moving modern changes we now recommend dotenvx. 

https://github.com/dotenvx/dotenvx

PR #838, opened 2024-07-26
Use "node:" prefix everywhere.

boenrobot:
Updated the minimum nodejs version and CI accordingly.

---

The use of the "node:" prefix enables use in runtimes like `workerd` where compatibility layer is provided for "node:" improts only, but not for unprefixed packages. The cost to that is that the minimum nodejs version needs to be raised to ">=14.18.0 <15 || >=16".

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,12 +12,12 @@ jobs:
 
     strategy:
       matrix:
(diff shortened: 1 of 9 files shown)


motdotla:
thank you but sorry not plans to bump the minimum version. many millions of projects depend on this on the older versions of node.

PR #842, opened 2024-08-27
fix: inline anchor groups

mceachen:
Looks like https://github.com/motdotla/dotenv/commit/dbb50ee902820cc0a188ab7a60ea08645372b293 replaced `(?:^|\A)` with `(?:^|^)`, and similarly, `(?:$|\z)` with `(?:$|$)`.

These non-capturing groups are equivalent to the simple symbol they wrap.

I thought maybe you were trying to work around an eslint or regex linter bug, but `eslint-plugin-regexp` passes both the prior and this new RegExp pattern.

Thanks for building dotenv!

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -6,7 +6,7 @@ const packageJson = require('../package.json')
 
 const version = packageJson.version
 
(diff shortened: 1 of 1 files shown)


motdotla:
i actually prefer them. `(?:^|^)` looks like a cat and `(?:$|$)` reminds me of the monopoly guy.

thank you for noticing though!

PR #845, opened 2024-11-14
Improved logging so when debug is true print added key values

larstelenor:
When debug is set to true a log statment will print out the value of  "${key}=${value}" was added to process.env

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -337,7 +337,7 @@ function populate (processEnv, parsed, options = {}) {
       }
     } else {
       processEnv[key] = parsed[key]
-      _debug(`"${key}" was added to process.env`)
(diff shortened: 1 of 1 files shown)


motdotla:
thank you for this @larstelenor but I'm going to leave out of dotenv. This opens up a pandora's box of - will these leak in my logs, how can i obfuscate the value, etc.

Instead of addressing all that here, we are addressing it in our more feature filled tool [dotenvx](https://github.com/dotenvx/dotenvx)

So if you want to debug values I recommend for anyone who reads this:

```
$ dotenvx run --debug -- yourcommand
```

It prints out a nice format of when values are pre-existing, overwritten, etc.

PR #858, opened 2025-03-06
docs: update install instructions

Electroid:
Adds bun alongside yarn in the install instructions. Also, removes `--save` install command, since that's been the default for years (since npm v5)

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,7 +58,7 @@ Dotenv is a zero-dependency module that loads environment variables from a `.env
 ## 🌱 Install
 
 ```bash
-npm install dotenv
+npm install dotenv --save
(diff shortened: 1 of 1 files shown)


motdotla:
looks good but can you add the --save back in. dotenv still has tons of people on old node and npm versions. i tend to keep things redundant sometimes with that in mind.

Electroid:
Sounds good, added `--save` back

motdotla:
thank you for the contribution!

motdotla:
also @Electroid have you guys seen [dotenvx](https://dotenvx.com) over at Bun? might find it interesting.

PR #863, opened 2025-04-10
chore(package): add homepage url

Fdawgs:
The addition of the [`homepage`](https://docs.npmjs.com/cli/v10/configuring-npm/package-json?v=true#homepage) property allows for the URL to the repo to be correctly displayed when hovered over in `package.json` in VS Code when using a proxy/private registry like in Azure DevOps:

Untitled

Compared to something like eslint-plugin-regexp, which [does have the homepage set](https://github.com/ota-meshi/eslint-plugin-regexp/blob/2d90a1aca863508647115c76a7b0c23b493eb0e0/package.json#L60):

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,6 +31,7 @@
     "type": "git",
     "url": "git://github.com/motdotla/dotenv.git"
   },
+  "homepage": "https://github.com/motdotla/dotenv#readme",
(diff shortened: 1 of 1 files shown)


motdotla:
thank you. this is great.

PR #867, opened 2025-06-13
feat: support nearest strategy to load .env from closest parent directory

### LINKED ISSUE #866: dotenv fails to load environment variables in nested directories on macOS Sequoia
## issue:

After updating to macOS Sequoia, dotenv no longer loads environment variables from the .env file when the script is executed from a nested directory. It only works when the script is in the project's root directory. Accessing process.env from a script in a subdirectory results in undefined for the expected variables.

## Reproduction:

Create a new project directory:

```Bash

mkdir my-project
cd my-project
Initialize a Node.js project:
```
```Bash

npm init -y
Install dotenv:
```
```Bash

npm install dotenv@16.5.0
Create a .env file in the root directory:

MY_VAR="Hello from the root"
Create a script in the root directory (index.js):
```
```JavaScript

require('dotenv').config();
console.log(process.env.MY_VAR);
Create a nested directory and a script inside it:
```
```Bash

mkdir src
```
```JavaScript

// src/nested.js
require('dotenv').config();
console.log(process.env.MY_VAR);
```
Run the scripts:

node index.js (from the root) -> Outputs: "Hello from the root" (Correct)
node src/nested.js (from the root) -> Outputs: undefined (Incorrect)

## Expected Behavior:

When running node src/nested.js, the output should be "Hello from the root", as dotenv should locate and load the .env file from the parent (root) directory.

Actual Behavior:

When running node src/nested.js, the output is undefined.

## Environment:

OS: macOS Sequoia
Device: Apple M1 Mac Mini
dotenv version: 16.5.0
Node.js version: 20.14.0
Notes:

This issue appears to be specific to the macOS Sequoia update. The same project structure and code worked as expected on previous versions of macOS. No error is thrown; process.env simply does not contain the variables from the .env file when the script is in a nested directory.

shyok21:
## What this PR does?

This PR adds support for a new config option: `{ strategy: 'nearest' }`  
When enabled, dotenv will look for the nearest `.env` file starting from `process.cwd()` and walking up the directory tree.  
This helps when working in monorepos or nested folder structures where multiple `.env` files might exist.

## Related Issue
Closes #866 

## Why it's useful

- Useful in monorepos where each package can have its own `.env`
- Prevents loading the wrong .env from the root when running scripts from a subfolder
- Backward-compatible: the default strategy remains unchanged if no config is passed

## How it works

- Added an optional `strategy: 'nearest'` to the config
- If strategy is set to 'nearest', the loader walks up from `process.cwd()` until it finds a `.env`
- If not found, behaves same as before

## Tests added

- Fixtures with nested folders and multiple .env files
- Tests for both default and 'nearest' strategy behavior

## Example usage

```js
require('dotenv').config({ strategy: 'nearest' })
```

### DIFF
--- a/lib/cli-options.js
+++ b/lib/cli-options.js
@@ -1,4 +1,4 @@
-const re = /^dotenv_config_(encoding|path|debug|override|DOTENV_KEY)=(.+)$/
+const re = /^dotenv_config_(encoding|path|debug|override|DOTENV_KEY|strategy)=(.+)$/
 
 module.exports = function optionMatcher (args) {
   return args.reduce(function (acc, cur) {
--- a/lib/env-options.js
+++ b/lib/env-options.js
@@ -21,4 +21,8 @@ if (process.env.DOTENV_CONFIG_DOTENV_KEY != null) {
   options.DOTENV_KEY = process.env.DOTENV_CONFIG_DOTENV_KEY
 }
 
+if (process.env.DOTENV_CONFIG_STRATEGY != null) {
+  options.strategy = process.env.DOTENV_CONFIG_STRATEGY
+}
+
 module.exports = options
--- a/lib/main.d.ts
+++ b/lib/main.d.ts
@@ -67,6 +67,15 @@ export interface DotenvConfigOptions {
   processEnv?: DotenvPopulateInput;
 
   /**
+   * Optional strategy to locate .env files from nearest parent directory upward.
+   *
+   * Default: `undefined`
+   *
+   * example: `require('dotenv').config({ strategy: 'nearest' })`
+   */
+  strategy?: 'nearest';
+
+  /**
    * Default: `undefined`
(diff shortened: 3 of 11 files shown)


motdotla:
Thank you for the very complete PR @shyok21 and the thoughtful description.

Regarding this:

> This helps when working in monorepos or nested folder structures where multiple .env files might exist.

Have you seen [dotenvx](https://github.com/dotenvx/dotenvx) ? It handles monorepos well as you can configure things to your specific needs.

I'll think on this PR you have here but I'm currently hesitant to add this to dotenv. Instead, I'm pushing people toward dotenvx for its added safety as well.

shyok21:
Thanks! Totally get your point. Just thought this opt-in config might help folks who prefer dotenv's simplicity in nested setups. Appreciate you considering it!

PR #868, opened 2025-06-16
fix: return first valid path in _vaultPath when array is provided

zegulas:
### 🔍 What This PR Does

This PR fixes a logic issue in the `_vaultPath()` function, which previously returned the **last valid match** from an array of paths in `options.path`, even if earlier paths were valid.

### ✅ Expected Behavior

Return the **first** existing `.env.vault` path when an array of paths is provided.

---

### 🔧 Summary of Changes

- Updated `_vaultPath` to return early on first valid path in array  
- Cleaned up unnecessary control flow in the function  
- Exported `_vaultPath` from `main.js` for testing  
- Added new test file: `tests/test-vault-path.js` to verify the fix  
- Ensured compliance with StandardJS linting rules  

---

### 🧪 Test Coverage

```yaml
npm test ✅ All tests pass
npm run lint ✅ Linting clean
```

---

### 🙌 Why It Matters

This makes behavior predictable and respects the order in which the user specifies fallback `.env` paths. It's a safer, more intuitive approach when supporting multiple environments.

Let me know if any updates are needed — happy to iterate!

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -156,24 +156,25 @@ function _instructions (result, dotenvKey) {
 }
 
 function _vaultPath (options) {
-  let possibleVaultPath = null
-
   if (options && options.path && options.path.length > 0) {
     if (Array.isArray(options.path)) {
       for (const filepath of options.path) {
-        if (fs.existsSync(filepath)) {
-          possibleVaultPath = filepath.endsWith('.vault') ? filepath : `${filepath}.vault`
+        const fullPath = filepath.endsWith('.vault') ? filepath : `${filepath}.vault`
+        if (fs.existsSync(fullPath)) {
+          return fullPath // Return first valid path
         }
       }
     } else {
-      possibleVaultPath = options.path.endsWith('.vault') ? options.path : `${options.path}.vault`
+      const fullPath = options.path.endsWith('.vault') ? options.path : `${options.path}.vault`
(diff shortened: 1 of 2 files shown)


motdotla:
@zegulas in what situation would you have more than one `.env.vault` file?

zegulas:
Thanks for the question, @motdotla.

You're right — I don't have a specific use case for multiple `.env.vault` files myself.

My intention with this PR was to address a **logical issue in the existing `_vaultPath()` function**. Since the function already accepts an array via `options.path`, I noticed that it loops through all the paths but ends up returning the **last valid match**, which feels counterintuitive and potentially error-prone.

If an array is allowed, it's generally expected to use the **first match**, not the last — especially since order implies priority. This change ensures that the function behaves consistently and predictably if/when such input is passed.

Let me know if you'd prefer to handle it differently or limit `options.path` to a single value. I'm happy to adjust the PR accordingly 🙏

motdotla:
let's just leave alone. I appreciate the thoughtfulness and especially noticing the difference in logic.

But I try to avoid adding logic unless a real world scenario merits it.

Also, the .env.vault mechanism is deprecated - replaced by [dotenvx](https://github.com/dotenvx/dotenvx) – so likely this either gets pulled out or replaced with dotenvx decryption in the future.

PR #869, opened 2025-06-24
Performance update

koraytuncer:


motdotla:
can you provide any stats showing this improved performance.

at first glance this seems largely unnecessary - especially considering .env files are generally small. it also looks ai generated rather than handmade.

koraytuncer:
There is no statistic showing this. I developed it with the support of artificial intelligence. Don't you use it?

motdotla:
yeah, AI is not going to do a good job at performance. And submitting a performance improvement without documenting the original performance vs the new is not something I can merge. Closed.

PR #874, opened 2025-06-27
default `quiet` to true

motdotla:


sefinek:
Why is `quiet` now set to `false` by default?

More:
https://github.com/motdotla/dotenv/issues/876#issuecomment-3015996776
https://github.com/sefinek/dotenv/issues/1

BoomchainLabs (commented):
Lgtm

dossy:
How do we specify `quiet: true` when importing as an ES6 module?

silverwind:
I disagree about this change. Libraries should not produce stdio logging by default.

At least I'm not affected because I only use the `parse` function which does to not produce this logging.

sefinek:
@silverwind, you see, some people just can't be talked out of certain things. #876

silverwind:
Hmm yeah I have drawn my conclusions already and have dropped this module in favor of [`util.parseEnv`](https://nodejs.org/api/util.html#utilparseenvcontent).

PR #891, opened 2025-07-10
Require process

cofenster-michal:
Require process instead of using global


motdotla:
I don't see why this is needed.

Issue #893, opened 2025-07-20
How do you set a custom config values for an es6 import?


andersr:
It appears the instructions only provide an example for how to set a custom config when using a commonjs import, ie 

```javascript 
require('dotenv').config({ foo: "bar" })
```

However, it is not clear how to do the same for an ES6 import. How do I set a custom config for this type of import? 
```js
import 'dotenv/config'
```

andersr:
Ok, figured it out:

```js
import dotenv from "dotenv";

dotenv.config({ path: "./.env.test" });
```

It might make sense to update the readme to include an es6 config example.

motdotla:
Ok yeah good point. Would you like to add that contribution? I can approve. Otherwise I will try and add it next week. Thanks!

andersr:
Yes, happy to add that. Please see #895

PR #895, opened 2025-07-21
Add info about ES6 import with config options

andersr:
Add an example of ES6 import with config option, to address issue raised in #893.


motdotla:
Thank you! 🙏

PR #915, opened 2025-10-07
fix: make DotenvPopulateInput accept NodeJS.ProcessEnv type

### LINKED ISSUE #767: DotenvPopulateInput not compatible with NodeJS.ProcessEnv
Seems that populate method just never worked in Typescript.
process.env may contain an undefined value in dict, making it unassignable.

```
import dotenv = require('dotenv')

const parsed = { HELLO: 'world' }
dotenv.populate(process.env, parsed)
```

> Argument of type 'ProcessEnv' is not assignable to parameter of type 'DotenvPopulateInput'.
>  'string' index signatures are incompatible.
>    Type 'string | undefined' is not assignable to type 'string'.
>      Type 'undefined' is not assignable to type 'string'.ts(2345)

dotenv type:
https://github.com/motdotla/dotenv/blob/cf4c56957974efb7238ecaba6f16e0afa895c194/lib/main.d.ts#L107C1-L109C2

node.js built-in type:
```
interface ProcessEnv extends Dict<string> {}

interface Dict<T> {
    [key: string]: T | undefined;
}
```

jedwards1211:
fix #767

### DIFF
--- a/lib/main.d.ts
+++ b/lib/main.d.ts
@@ -124,7 +124,7 @@ export interface DotenvPopulateOptions {
 }
 
 export interface DotenvPopulateInput {
-  [name: string]: string;
+  [name: string]: string | undefined;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


motdotla:
thank you! going out shortly.

PR #919, opened 2025-10-23
Remove Ads

### LINKED ISSUE #918: Why is this library muddying my logs with ads
The audacity of the maintainers of this library to think they can just throw ads in my server.

Example log output on startup.
```js
[dotenv@17.2.3] injecting env (0) from .env -- tip: 📡 add observability to secrets: https://dotenvx.com/ops
```

thecoolwinter:
Removed the logging of ads from the library.

closes #918

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -6,31 +6,6 @@ const packageJson = require('../package.json')
 
 const version = packageJson.version
 
-// Array of tips to display randomly
-const TIPS = [
-  '🔐 encrypt with Dotenvx: https://dotenvx.com',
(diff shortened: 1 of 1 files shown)


thecoolwinter:
Users should not have to add a brand new config option to avoid annoying ads for your new service. Seriously consider removing them. A far less invasive route would be through an install message, but this is annoying and just not cool generally.

motdotla:
this has been discussed at length. thank you the PR but closing it. things will stay as I've intended. all the best.

ghost:
@motdotla 
5df6608-1

PR #931, opened 2026-02-12
Add explicit test for duplicate keys in parse (last value wins)

codehassoul:
Adds an explicit test to document behavior when duplicate keys
appear within a single .env file. Currently, the last occurrence
wins due to object assignment order. This test formalizes that
behavior as part of the parse contract.


motdotla:
great contribution. thank you for this!

PR #938, opened 2026-02-25
fix: route _log/_debug to stderr, simplify LINE regex, add troubleshooting docs

sudip-kumar-prasad:
## Summary

This PR addresses three independent improvements to the dotenv codebase.

---

### Fix 1: Route [_log](cci:1://file:///Users/sudipkumarprasad/Desktop/dotenv/lib/main.js:137:0-139:1) and [_debug](cci:1://file:///Users/sudipkumarprasad/Desktop/dotenv/lib/main.js:133:0-135:1) output to `stderr`

**Problem:** [_log()](cci:1://file:///Users/sudipkumarprasad/Desktop/dotenv/lib/main.js:137:0-139:1) and [_debug()](cci:1://file:///Users/sudipkumarprasad/Desktop/dotenv/lib/main.js:133:0-135:1) both used `console.log`, which writes to `stdout`. This causes diagnostic/status messages to pollute the standard output stream, breaking common shell patterns like:

```sh
node app.js | jq         # dotenv tip messages get mixed into the JSON stream
node app.js > output.txt # log messages end up written to the file

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -43,7 +43,7 @@ function dim (text) {
   return supportsAnsi() ? `\x1b[2m${text}\x1b[0m` : text
 }
 
-const LINE = /(?:^|^)\s*(?:export\s+)?([\w.-]+)(?:\s*=\s*?|:\s+?)(\s*'(?:\\'|[^'])*'|\s*"(?:\\"|[^"])*"|\s*`(?:\\`|[^`])*`|[^#\r\n]+)?\s*(?:#.*)?(?:$|$)/mg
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: lib/main.js [hunk 1]
```diff
@@ -43,7 +43,7 @@ function dim (text) {
   return supportsAnsi() ? `\x1b[2m${text}\x1b[0m` : text
 }
 
-const LINE = /(?:^|^)\s*(?:export\s+)?([\w.-]+)(?:\s*=\s*?|:\s+?)(\s*'(?:\\'|[^'])*'|\s*"(?:\\"|[^"])*"|\s*`(?:\\`|[^`])*`|[^#\r\n]+)?\s*(?:#.*)?(?:$|$)/mg
```
motdotla: i like this part of the regex - because i like that it looks like a cat. i've mentioned this before and plan to keep it.

motdotla:
not significant enough of a change. doesn't add much value and could break things given historical artifact.

PR #941, opened 2026-03-15
fix: write log and debug output to stderr instead of stdout

### LINKED ISSUE #903: Advertisements or whatever for dotenvx should go to stderr not stdout
Hi, 

While I admit I'm not a huge fan of these advertisements being in the lib, I do get what your going for. However, out putting into stdout breaks piping and things, the proper place really should be stderr.  

Because it's in standard out, I have to go back and add quiet to all the places where it's used that pipes to other scripts and things. Stderr is generally where warnings etc... are, and isn't typically used for structured data piping.

Cheers and good luck with dotenvx

mahmoodhamdi:
## What

Changed `_log()` and `_debug()` functions in `lib/main.js` to write to `stderr` (`console.error`) instead of `stdout` (`console.log`).

## Why

dotenv's informational log messages (injection summaries, tips, debug output) currently go to `stdout`, which breaks:

- **Piping** — scripts that pipe stdout to other tools get dotenv's messages mixed into structured data
- **MCP servers** — JSON-RPC communication over stdio fails because dotenv's log lines are not valid JSON (reported in the comments by multiple MCP server maintainers)
- **Any stdout-dependent protocol** — tools that read structured output from stdout get unexpected lines

`stderr` is the correct destination for diagnostic/informational messages per Unix convention. The `_warn()` function already correctly uses `console.error` — this PR makes `_log()` and `_debug()` consistent with that pattern.

## Changes

- `lib/main.js`: `_log()` and `_debug()` now use `console.error()` instead of `console.log()`
- Tests updated to stub `console.error` instead of `console.log` where testing log/debug output

## Tests

All 194 existing tests pass. No new tests needed — this is a one-line behavioral change in two functions.

Fixes #903

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -132,11 +132,11 @@ function _warn (message) {
 }
 
 function _debug (message) {
-  console.log(`[dotenv@${version}][DEBUG] ${message}`)
+  console.error(`[dotenv@${version}][DEBUG] ${message}`)
 }
 
 function _log (message) {
-  console.log(`[dotenv@${version}] ${message}`)
+  console.error(`[dotenv@${version}] ${message}`)
 }
 
 function _dotenvKey (options) {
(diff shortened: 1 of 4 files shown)


mahmoodhamdi:

PR #944, opened 2026-03-15
fix: export DotenvError type for typed error handling

mahmoodhamdi:
## What

Added `export` keyword to the `DotenvError` type in `lib/main.d.ts` and added a type test.

## Why

`DotenvError` is defined in the type declarations and used by `DotenvConfigOutput.error`, but it was not exported. This prevents TypeScript consumers from properly typing error handling:

```typescript
import { config, DotenvError } from 'dotenv';

const result = config();
if (result.error) {
  // Before: no way to reference DotenvError type
  // After: can use DotenvError["code"] for typed error codes
  const code: DotenvError["code"] = result.error.code;
  
  switch (code) {
    case 'MISSING_DATA':
    case 'INVALID_DOTENV_KEY':
    case 'DECRYPTION_FAILED':
      // handle specific error codes with type safety
      break;
  }
}
```

## Changes

- `lib/main.d.ts`: Added `export` to `type DotenvError`
- `tests/types/test.ts`: Added type test importing and using `DotenvError`

## Tests

- All 194 tests pass
- TypeScript type check passes (`npm run dts-check`)

### DIFF
--- a/lib/main.d.ts
+++ b/lib/main.d.ts
@@ -94,7 +94,7 @@ export interface DotenvConfigOutput {
   parsed?: DotenvParseOutput;
 }
 
-type DotenvError = Error & {
+export type DotenvError = Error & {
   code: 
     | 'MISSING_DATA'
     | 'INVALID_DOTENV_KEY'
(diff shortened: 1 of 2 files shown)


mahmoodhamdi:

PR #946, opened 2026-03-15
docs: fix function count and remove incorrect parse options

mahmoodhamdi:
## What

Two fixes in the Docs section of README.md:

1. **Fixed function count**: The docs stated "Dotenv exposes four functions" but only listed three (`config`, `parse`, `populate`). Updated to list all five public functions including `configDotenv` and `decrypt`.

2. **Removed incorrect parse options**: The Parse section documented a `debug` option with an example, but `parse()` only accepts a single `src` parameter — it has no options argument. The TypeScript declaration confirms this:
   ```typescript
   export function parse<T>(src: string | Buffer): T;
   ```

## Tests

All 194 tests pass.

### DIFF
--- a/README.md
+++ b/README.md
@@ -578,11 +578,13 @@ Alternatively, just use [dotenv-webpack](https://github.com/mrsteele/dotenv-webp
 
 ## Docs
 
-Dotenv exposes four functions:
+Dotenv exposes five functions:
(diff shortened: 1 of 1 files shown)


mahmoodhamdi:

PR #947, opened 2026-03-15
test: add tests for populate() return value

mahmoodhamdi:
## What

Added 3 new test cases for the `populate()` function's return value in `tests/test-populate.js`.

## Why

The `populate()` function returns an object containing only the keys that were actually set (used by `configDotenv` for the injection count in log messages). This return behavior had no test coverage.

## New Tests

| Test | Scenario | Expected Return |
|------|----------|----------------|
| All new keys | Empty target, two new keys | Both keys returned |
| Existing key without override | One existing key, one new key | Only new key returned |
| Existing key with override | One existing key, one new key, `override: true` | Both keys returned |

## Tests

All 200 tests pass (194 existing + 6 new, 3 in this PR + 3 already existed in previous run).

### DIFF
--- a/tests/test-populate.js
+++ b/tests/test-populate.js
@@ -89,6 +89,42 @@ t.test('logs populating when debug mode and override turned on', ct => {
   logStub.restore()
 })
 
+t.test('returns object with only newly set keys', ct => {
+  ct.plan(2)
+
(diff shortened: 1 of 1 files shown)


mahmoodhamdi:

PR #950, opened 2026-03-15
docs: fix function count and remove incorrect parse options in README-es

mahmoodhamdi:
## Summary
- Fix "four functions" to "five functions" in the Spanish README, listing `config`, `configDotenv`, `parse`, `populate`, and `decrypt`
- Remove incorrect `Options` section under `Parse` that documented a `debug` option — `parse()` only accepts `src` (String or Buffer), not options
- Same fix as #946 but for `README-es.md`

## Test plan
- [x] Verified `parse()` signature in source: only accepts `src` parameter

### DIFF
--- a/README-es.md
+++ b/README-es.md
@@ -578,11 +578,13 @@ Alternatively, just use [dotenv-webpack](https://github.com/mrsteele/dotenv-webp
 
 ## Docs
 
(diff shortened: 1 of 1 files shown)


mahmoodhamdi:

PR #955, opened 2026-03-15
chore: remove duplicate .env-multiline test fixture

mahmoodhamdi:
## Summary
- Remove `tests/.env-multiline` (with hyphen) which is an identical duplicate of `tests/.env.multiline` (with dot)
- Only the dotted version (`tests/.env.multiline`) is referenced by `test-parse-multiline.js`
- The hyphenated version is unused and was likely left over from a rename

## Test plan
- [x] Verified no code references `tests/.env-multiline` (grep confirms only `.env.multiline` is used)


mahmoodhamdi:

PR #957, opened 2026-03-15
fix: remove dead individual exports overwritten by module.exports assignment

mahmoodhamdi:
## Summary
- Remove 7 individual `module.exports.X = ...` property assignments that were dead code
- These assignments on lines 423-429 were immediately overwritten by `module.exports = DotenvModule` on line 431
- The `DotenvModule` object already contains all the same properties, so the final assignment is sufficient

## Test plan
- [x] Verified `module.exports = DotenvModule` exports all functions correctly (configDotenv, _configVault, _parseVault, config, decrypt, parse, populate)

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -420,12 +420,4 @@ const DotenvModule = {
   populate
 }
 
-module.exports.configDotenv = DotenvModule.configDotenv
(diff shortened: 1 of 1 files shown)


mahmoodhamdi:

PR #958, opened 2026-03-15
test: add env-options edge case tests

mahmoodhamdi:
## Summary
- Add tests for `lib/env-options.js` edge cases not previously covered
- Empty string values are included (the `!= null` check passes for empty strings)
- Boolean options (`quiet`, `debug`, `override`) set to `'false'` string
- Multiple options set simultaneously

## Test plan
- [x] All tests pass (`npm test` — 199 pass)


mahmoodhamdi:

PR #959, opened 2026-03-15
test: verify vault missing warning uses console.error

mahmoodhamdi:
## Summary
- Fix vault warning tests to actually verify the `_warn()` call via `console.error`
- Previously, these tests only stubbed `console.log` and passed because the `configDotenv` fallback calls `_log()` (which uses `console.log`), not because the warning was verified
- The `_warn` function has always used `console.error`, but the tests never checked it

## Test plan
- [x] All tests pass (`npm test` — 196 pass)
- [x] New assertions verify `console.error` is called when vault file is missing


mahmoodhamdi:

PR #960, opened 2026-03-15
test: add multi-path partial failure tests

mahmoodhamdi:
## Summary
- Add tests for `config()` behavior when using array of paths where some files exist and others don't
- Verify that parsed values are returned from valid files even when other files in the array fail
- Verify that the error from the failed file is returned alongside the parsed data
- Tests both orderings: valid file first, and invalid file first

## Test plan
- [x] All tests pass (`npm test` — 200 pass)


mahmoodhamdi:

PR #961, opened 2026-03-15
test: add DOTENV_KEY precedence tests

mahmoodhamdi:
## Summary
- Add test verifying `options.DOTENV_KEY` takes precedence over `process.env.DOTENV_KEY`
- Add test verifying `process.env.DOTENV_KEY` is used as fallback when option is not set
- Documents the expected priority: options > env var > empty string

## Test plan
- [x] All tests pass (`npm test` — 198 pass)


mahmoodhamdi:

PR #970, opened 2026-03-15
test: add tests for populate() return value with partial keys

mahmoodhamdi:
## Summary

- Add test verifying `populate()` returns only keys that were actually set (skipped keys excluded)
- Add test verifying `populate()` returns all keys when `override: true`

The `populate()` function returns a `populated` object tracking which keys were set, but this return value had no test coverage.

## Test plan

- [x] `npm test` passes (198 tests)
- [x] No new dependencies


mahmoodhamdi:

PR #972, opened 2026-03-15
test: verify DOTENV_CONFIG_QUIET from .env file suppresses log

mahmoodhamdi:
## Summary

- Add test verifying that `DOTENV_CONFIG_QUIET=true` set inside the `.env` file itself properly suppresses the log message
- Add `.env.quiet` test fixture with `DOTENV_CONFIG_QUIET=true`

This tests the re-read logic in `configDotenv` (lines 298-300) where `DOTENV_CONFIG_QUIET` is re-evaluated from processEnv after population, allowing users to set it inside their `.env` file.

## Test plan

- [x] `npm test` passes (196 tests)
- [x] No new dependencies

### DIFF
--- a/tests/.env.quiet
+++ b/tests/.env.quiet
@@ -1,0 +1,2 @@
+DOTENV_CONFIG_QUIET=true
+HELLO=world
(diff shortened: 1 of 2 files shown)


mahmoodhamdi:

PR #973, opened 2026-03-15
test: add tilde home directory expansion test for array paths

mahmoodhamdi:
## Summary

- Add test verifying `~` home directory expansion works correctly when `path` option is an array (e.g. `{ path: ['~/.env.local', '~/.env'] }`)

The existing `~` expansion test only covers single path strings. Array paths go through the same `_resolveHome` function but via a different code branch.

## Test plan

- [x] `npm test` passes (197 tests)
- [x] No new dependencies

### DIFF
--- a/tests/test-config.js
+++ b/tests/test-config.js
@@ -108,6 +108,22 @@ t.test('takes option for path along with home directory char ~', ct => {
   ct.end()
 })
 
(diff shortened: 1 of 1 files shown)


mahmoodhamdi:

PR #976, opened 2026-03-15
test: add decrypt output parsing and type tests

mahmoodhamdi:
## Summary

- Add test verifying `decrypt()` output can be parsed by `parse()` (roundtrip test)
- Add test verifying `decrypt()` returns a string type

The existing test only checks the raw decrypted string value.

## Test plan

- [x] `npm test` passes (196 tests)
- [x] No new dependencies


mahmoodhamdi:

PR #978, opened 2026-03-15
test: add \\r carriage return expansion tests for parse

mahmoodhamdi:
## Summary

- Add test verifying `\r` is expanded to carriage return in double-quoted values
- Add test verifying `\r` is NOT expanded in unquoted values
- Add test verifying `\r` is NOT expanded in single-quoted values

The parser expands both `\n` and `\r` in double-quoted values (lines 76-77 of main.js), but only `\n` expansion had test coverage.

## Test plan

- [x] `npm test` passes (197 tests)
- [x] No new dependencies


mahmoodhamdi:

PR #980, opened 2026-03-15
test: verify log message format includes version and injecting info

mahmoodhamdi:
Adds a test to confirm that the log message emitted by `config()` includes:

- The version prefix `[dotenv@`
- The descriptive text `injecting env`

This ensures the log output format remains consistent with user-facing documentation and expectations.


mahmoodhamdi:

PR #981, opened 2026-03-15
test: verify log message shows correct populated key count

mahmoodhamdi:
Adds a test confirming that the log message emitted by `config()` includes the correct count of populated keys in the `(N)` format, ensuring the count shown to users accurately reflects how many environment variables were injected.


mahmoodhamdi:

PR #982, opened 2026-03-15
test: add parse tests for escaped quotes inside same-type quotes

mahmoodhamdi:
Adds tests verifying the behavior of backslash-escaped quotes within their own quote type:

- `\"` inside double-quoted values
- `\'` inside single-quoted values
- `` \` `` inside backtick-quoted values

Confirms that the parser correctly uses escaping to find the closing quote delimiter while retaining the backslash characters in the parsed output.


mahmoodhamdi:

PR #984, opened 2026-03-15
test: verify warn message uses console.error with [WARN] prefix

mahmoodhamdi:
Adds a test confirming that warning messages (from `_warn`) are correctly sent to `console.error` (stderr) rather than `console.log` (stdout), and that they include the `[WARN]` prefix in the message format.

This is important for users piping dotenv output, as warnings should go to stderr to avoid interfering with stdout data.


mahmoodhamdi:

PR #985, opened 2026-03-15
test: verify processEnv null defaults to process.env

mahmoodhamdi:
Adds a test confirming that passing `processEnv: null` falls back to writing environment variables to `process.env`, verifying the `!= null` check behavior in `configDotenv`.


mahmoodhamdi:

PR #987, opened 2026-03-15
test: add parse tests for keys without values

mahmoodhamdi:
Adds tests for parse edge cases involving keys without values:

- Key without an equals sign is ignored (not included in parsed output)
- Key with a trailing equals sign (`KEY=`) has an empty string value
- Key with equals and trailing whitespace (`KEY= `) has an empty string value


mahmoodhamdi:

PR #988, opened 2026-03-15
test: add debug encoding message tests

mahmoodhamdi:
Adds tests verifying the debug mode encoding behavior:

- When no encoding is specified, debug mode logs a message about UTF-8 being the default
- When encoding is explicitly provided, the UTF-8 default message is not logged

This covers the encoding branch in `configDotenv` that was previously untested.


mahmoodhamdi:

PR #989, opened 2026-03-15
test: add parse tests for special characters in values

mahmoodhamdi:
Adds tests verifying that `parse()` correctly handles common real-world values containing special characters:

- URLs with query parameters (e.g., database connection strings)
- Base64-encoded values with trailing equals signs
- Values with semicolons (e.g., PATH-like variables)


mahmoodhamdi:

PR #990, opened 2026-03-15
test: add populate tests for empty parsed and non-overlapping keys

mahmoodhamdi:
Adds tests for `populate()` edge cases:

- Returns empty object when parsed input is empty (no keys to set)
- Does not modify existing processEnv keys that are not present in parsed
- Only adds new keys from parsed to processEnv


mahmoodhamdi:

PR #991, opened 2026-03-15
test: verify log message shows relative path to .env file

mahmoodhamdi:
Adds a test confirming that the log message emitted by `config()` includes the relative path to the `.env` file, ensuring users see helpful path information in the output.


mahmoodhamdi:

PR #992, opened 2026-03-15
test: add multi-path with override test for process.env

mahmoodhamdi:
Adds a test verifying that `config()` with multiple paths and `override: true` correctly:

- Writes values to `process.env` even when keys already exist
- Uses the second file's values when both files define the same key (since override applies to the internal `parsedAll` aggregation too)


mahmoodhamdi:

PR #993, opened 2026-03-15
test: add decrypt error tests for short key and wrong key

mahmoodhamdi:
Adds tests verifying `decrypt()` error handling:

- Throws `INVALID_DOTENV_KEY` for keys shorter than 64 characters
- Throws `DECRYPTION_FAILED` when the key doesn't match the ciphertext


mahmoodhamdi:

PR #995, opened 2026-03-15
test: verify configDotenv returns same result as config

mahmoodhamdi:
Adds a test confirming that `configDotenv()` and `config()` produce identical results when no `DOTENV_KEY` is set, verifying the documented equivalence of these two exported functions.


mahmoodhamdi:

PR #997, opened 2026-03-23
feat: expand \t escape sequence in double-quoted values

mahmoodhamdi:
## What

Adds support for expanding `\t` (tab) escape sequences inside double-quoted values in `.env` files, matching the existing behavior for `\n` and `\r`.

## Why

I use dotenv in all my Node.js projects. I recently needed to set a value containing tabs (a TSV-formatted string) in my `.env` file. I naturally wrote:

```
TSV_HEADER="name\tage\temail"
```

I was surprised when the literal `\t` came through instead of actual tab characters, since `\n` and `\r` already work as expected in double-quoted values. This is also consistent with how shell double-quoting handles escape sequences.

## How

Added a single line in `parse()` to replace `\t` with actual tab characters when the value is double-quoted — the same approach already used for `\n` and `\r`.

## Tests

- All 194 existing tests still pass
- Added 3 new test cases:
  - `\t` is expanded in double-quoted values
  - `\t` is **not** expanded in unquoted values
  - `\t` is **not** expanded in single-quoted values


mahmoodhamdi:

PR #998, opened 2026-03-23
fix: use DotenvPopulateOptions for populate() third argument type

mahmoodhamdi:
## Problem

The `populate()` function accepts three arguments: `processEnv`, `parsed`, and `options`. Its implementation only reads `debug` and `override` from `options` — exactly the two fields defined in `DotenvPopulateOptions`.

However, the type signature in `lib/main.d.ts` typed the third argument as `DotenvConfigOptions`, which is the broader interface that also includes `path`, `encoding`, `quiet`, `processEnv`, and `DOTENV_KEY`. None of those extra fields do anything inside `populate()`.

This meant TypeScript users got misleading IntelliSense: autocomplete would suggest `quiet`, `path`, `encoding`, etc. as valid options for `populate()`, when they are silently ignored at runtime.

There was also a secondary issue: the JSDoc `@param options` example on `populate()` showed `{ quiet: false, debug: true, override: false }`, referencing `quiet` which is not part of `DotenvPopulateOptions` and does nothing in `populate()`.

## Changes

- **`lib/main.d.ts`**: Change `populate()` third argument from `DotenvConfigOptions` to `DotenvPopulateOptions`
- **`lib/main.d.ts`**: Fix `@param options` JSDoc example to remove `quiet` and only show `debug`/`override`
- **`lib/main.d.ts`**: Fix the JSDoc examples inside `DotenvPopulateOptions` fields to reference `populate()` instead of `config()`
- **`tests/types/test.ts`**: Add `populate()` usage to the TypeScript type test so the `dts-check` script covers this function going forward

## Verification

```
npm test  # all 194 tests pass, lint clean, dts-check passes
```

The `DotenvPopulateOptions` interface already existed in the type file — it just wasn't being used for `populate()`. This change connects the two correctly.

### DIFF
--- a/lib/main.d.ts
+++ b/lib/main.d.ts
@@ -95,7 +95,7 @@ export interface DotenvConfigOutput {
 }
 
 type DotenvError = Error & {
-  code: 
+  code:
     | 'MISSING_DATA'
     | 'INVALID_DOTENV_KEY'
     | 'NOT_FOUND_DOTENV_ENVIRONMENT'
@@ -109,7 +109,7 @@ export interface DotenvPopulateOptions {
    *
    * Turn on logging to help debug why certain keys or values are not being set as you expect.
    *
-   * example: `require('dotenv').config({ debug: process.env.DEBUG })`
+   * example: `require('dotenv').populate(processEnv, parsed, { debug: true })`
    */
   debug?: boolean;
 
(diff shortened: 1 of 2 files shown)


motdotla:
thank you for a helpful contribution.

PR #1010, opened 2026-04-15
perf: ~2x faster parse() via hand-written character scanner

homanp:
## Summary

Swaps the `LINE` regex in `parse()` for a hand-rolled character scanner with a `Uint8Array(256)` lookup table for key chars. About 2x faster on real .env files. Same behavior, same API, still zero deps.

## Benchmarks

Repro:

```
node tests/test-parse-perf.js
```

5000 parses of a 200-line .env, median of 7 runs, three fresh node processes per side (Node 23, macOS):

|  | run 1 | run 2 | run 3 |
|---|---|---|---|
| master | 121.21 ms | 121.77 ms | 122.85 ms |
| this PR | 57.86 ms | 58.71 ms | 58.26 ms |

~2.1x speedup, variance under 2% on both sides. Also tried it against a realistic ~50-key production-style .env (db, auth, aws, flags) — similar story, roughly 1.9x.

## What changed

The regex was the bottleneck. Replaced it with a single pass over the string using `charCodeAt`:

- `Uint8Array(256)` built at module load tells us in O(1) whether a byte is a valid key char (`[A-Za-z0-9._-]`).
- For unquoted values, `indexOf` finds the `\n` / `#` boundaries — V8's `indexOf` is way faster than walking char by char.
- CR normalization only runs when there's actually a `\r` in the input, instead of always.

## Tests

Existing test suite passes. I also diffed output against master across a bunch of edge cases I was worried about and got byte-identical results: escaped quotes inside double quotes, unterminated quotes, `KEY: value` (yaml-style), `export KEY=...`, multiline values with `\n` escapes, inline comments after quoted values, `KEY==value`, `KEY=val#notcomment`, `KEY=#nothing`, leading/trailing whitespace, `\r` line endings, empty input, comment-only input.

Happy to add any of those as actual TAP cases if you'd like them in the suite.

## Notes

No API changes, no behavior changes, no new deps.

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -35,41 +35,153 @@ function dim (text) {
   return supportsAnsi() ? `\x1b[2m${text}\x1b[0m` : text
 }
 
-const LINE = /(?:^|^)\s*(?:export\s+)?([\w.-]+)(?:\s*=\s*?|:\s+?)(\s*'(?:\\'|[^'])*'|\s*"(?:\\"|[^"])*"|\s*`(?:\\`|[^`])*`|[^#\r\n]+)?\s*(?:#.*)?(?:$|$)/mg
+const KEY_CHAR = new Uint8Array(256)
+for (let _i = 48; _i <= 57; _i++) KEY_CHAR[_i] = 1
+for (let _i = 65; _i <= 90; _i++) KEY_CHAR[_i] = 1
+for (let _i = 97; _i <= 122; _i++) KEY_CHAR[_i] = 1
+KEY_CHAR[45] = 1; KEY_CHAR[46] = 1; KEY_CHAR[95] = 1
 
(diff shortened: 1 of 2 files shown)


motdotla:
@homanp wow, one of the most interesting contributions to dotenv in a decade.

Give me time on this one to consider the maintenance cost here. It's much less readable, but it is very interesting.

homanp:
> @homanp wow, one of the most interesting contributions to dotenv in a decade.
> 
> 
> 
> Give me time on this one to consider the maintenance cost here. It's much less readable, but it is very interesting.
> 
> 
> 
> 

Thank you, happy to contribute. I have some other ideas as well for cold boot but haven't tested them yet.

Open to feedback on this PR. Happy to make optimizations based on your thoughts

PR #1015, opened 2026-04-21
Set lead and maintainer logins to fork owner

alanzabihi:
PROGRAM.md still had `motdotla` (upstream owner) as `lead_github_login` and `maintainer_github_login`. This caused `polyresearch lead` to reject with a lead-only guard error since the fork's GitHub login is `alanzabihi`.


alanzabihi:
Opened against wrong repo — should target the fork.

PR #1025, opened 2026-06-18
docs: add Deno to install instructions

bartlomieju:
👋 Bartek from the Deno team here. Thanks for dotenv.

The Advanced section has per-package-manager `<details>` blocks for bun / yarn / pnpm. This adds a matching `deno` one, since Deno is a drop-in replacement for npm these days:

```sh
deno add dotenv
```

Docs-only change that mirrors the existing blocks. Totally fine to close this if you'd prefer not to. Happy to adjust.

_Disclosure: I work on Deno, so I have an interest here. The goal is parity with the package managers you already list, not promotion. This PR was prepared with AI assistance under human supervision: I review every change myself and personally respond to any comments or review feedback._

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,6 +99,13 @@ pnpm add dotenv
 ```
 
 </details>
+<details><summary>deno</summary><br>
+
+```sh
+deno add dotenv
+```
+
+</details>
 <details><summary>Monorepos</summary><br>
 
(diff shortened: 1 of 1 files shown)


motdotla:
thanks for this but yeah I think it is unnecessary detail.

motdotla:
actually, changed my mind. merging.

PR #1029, opened 2026-06-22
Enhance .env parsing for inline comments with leading whitespace

### LINKED ISSUE #1018: Incomplete Parsing Caused by # Symbol in Configuration Content
Hello, I've found a bug in the configuration reading function.
When the configuration content contains characters such as test*&#, the # will be treated as a comment delimiter after being read, making it impossible to obtain the complete configuration content normally.

udithavithanage:
This pull request improves the parsing of inline comments in environment variable files, especially around handling the `#` character in values. The main change is to correctly distinguish between `#` as part of a value and `#` as the start of a comment, depending on whether it is preceded by whitespace. This is reflected in both the parser logic and expanded test coverage.

## Parsing improvements

- Updated the `LINE` regular expression in `lib/main.js` to treat `#` as the start of a comment only when it is preceded by whitespace.
- This allows `#` to be included as part of an environment variable value when it is not preceded by whitespace.

## Examples

| Input                       | Output               |
| --------------------------- | -------------------- |
| `DB_PASS=test*&#secret123`  | `test*&#secret123` |
| `DB_PASS=test #secret123`   | `test`|
| `DB_PASS="test #secret123"` | `test #secret123`|

## Test coverage enhancements

- Added new test cases in `tests/.env`, `tests/.env.multiline`, and corresponding assertions in `tests/test-parse.js` and `tests/test-parse-multiline.js`.
- Added coverage for:
  - Inline comments following values.
  - `#` characters embedded in unquoted values.
  - `#` characters inside quoted values.
  - Cases where `#` is and is not preceded by whitespace.

These tests verify that comments are parsed correctly while preserving `#` characters that are intended to be part of the value.

Fixes: #1018


udithavithanage:
Hi @motdotla , thanks for your time! Could you please review this PR and merge it if it looks good? 🙌

motdotla:
thank you for the contribution @udithavithanage but this would be a MAJOR breaking change.

dotenv respects anything after # as a comment. dotenvx does the same and most other dotenv libraries in different languages as well. it's a bit unintuitive admittedly but it is so embedded now that a change like this could wreck millions or maybe even hundreds of millions of .env files

if you need to use a # inside a value just use double quotes.

Issue #1043, opened 2026-08-23
fast parser diverges from the default parser (silent value loss)


chuanghiduoc:
---
name: fast parser diverges from the default parser (silent value loss with `{ fast: true }`)
about: parse(src, {fast:true}) returns different values than parse(src) for several inputs — including silently dropping values
title: fast parser diverges from the default parser (silent value loss)
labels: bug
assignees: ''

---

The opt-in fast parser (`parse(src, { fast: true })`, merged in #1010) disagrees with the default regex parser on several classes of input. Since `fast` is meant as a drop-in performance option, any divergence means users who flip the flag get different env values — in one case a value is silently replaced with an empty string.

Reproduced on master (after #1041 and #1042), Node 24.12.0, Windows:

### 1. Blank line before a quoted multiline value — value lost

```js
require('dotenv').parse('A=\n\n"hello\nworld"')
// default: { A: 'hello\nworld' }
require('dotenv').parse('A=\n\n"hello\nworld"', { fast: true })
// fast:   { A: '' }        <- secret silently becomes empty string
```

The default `LINE` regex lets `\s*` inside the quoted-value alternative span newlines, so a quote on a later line is consumed as A's value. The fast scanner ends A at end-of-line and then fails to parse `"hello\nworld"` as a key line.

This also fires without blank lines: `A=\n'b'` gives `{A: 'b'}` vs `{A: ''}`.

### 2. Form feed / vertical tab / non-breaking space before a key — key dropped

```js
require('dotenv').parse('\fA=1')            // { A: '1' }
require('dotenv').parse('\fA=1', {fast: true}) // {}
// same for \v and U+00A0
```

The default parser skips these because `\s` matches them; the fast scanner only skips space/tab/newline/BOM, so when the key scan finds no `[A-Za-z0-9_.-]` it discards the whole line. NBSP is common in .env files edited via copy-paste.

### 3. Trailing junk after a closing quote — different value

```js
require('dotenv').parse('TOKEN="abc" oops')
// default: { TOKEN: '"abc" oops' }
require('dotenv').parse('TOKEN="abc" oops', { fast: true })
// fast:   { TOKEN: 'abc' }
```

Default keeps everything after `=` (the junk falls outside the match), fast returns the clean quoted content. Arguably fast's result is nicer here, but it differs.

---

Tests demonstrating all five failures (`tests/test-fast-parity.js`, tap):

```
not ok 2 - fast must match default            (blank line case)
[...]

PR #1044, opened 2026-08-24
fix: align the fast parser with the default parser

chuanghiduoc:
Follow-up to #1043. Aligns the fast parser with the default parser on all three divergence classes from the issue.

## Changes

**1. Quoted value starting on a later line after `KEY=`**

The default regex's quoted alternative carries a greedy leading `\s*` that crosses newlines, so a quote on a later line is consumed as the value. The scanner now detects when a newline was crossed after the separator and, if the next non-blank char is a quote, parses the value with full multiline support:

```js
require('dotenv').parse('A=\n\n"hello\nworld"', { fast: true })
// before: { A: '' }
// after:  { A: 'hello\nworld' }   (matches default)
```

Junk after the closing quote on that line still kills the whole alternative (the raw fallback cannot cross newlines), matching `A=\n\n"x" junk` → `{ A: '' }`.

**2. `\f`, `\v`, U+00A0 before a key**

Added to the whitespace-skip set alongside space/tab/newline/BOM:

```js
parse('\u00A0A=1', { fast: true })
// before: {}
// after:  { A: '1' }
```

**3. Trailing junk after a closing quote**

`TOKEN="abc" oops` now keeps the raw single-line value `"abc" oops` (the quoted form only survives when nothing but spaces/tabs or a comment follows the closing quote — anything else makes the raw `[^\r\n#]+` fallback take over starting at the opening quote, exactly like the regex backtracks).

## Tests

- `tests/test-fast-parity.js` — parity assertions for the three issue cases; failed on master (`5 fail`), passes now.
- Existing `test-parse-fast.js` cross-checks against `tests/.env` and the multiline fixture: failed on master, passes with this change.

Full suite on Windows: `201 total / 197 pass / 3 skip-adjacent fails` — the 3 remaining failures (`deals with file:// path`, `displays the injected env message without tips`) are pre-existing environment failures that also fail on unmodified master.


motdotla:
fixing here and will go out with next patch: https://github.com/motdotla/dotenv/pull/1056

PR #1046, opened 2026-08-29
fix: keep fast parser behavior in parity with classic parser

### LINKED ISSUE #1043: fast parser diverges from the default parser (silent value loss)

amtbsl:
## Summary

- fall back to the classic parser for JavaScript whitespace cases the fast path cannot safely handle
- preserve quoted values that begin after a newline
- treat junk after a closing quote like the classic parser
- add focused parser parity regressions

## Testing

- `npm test` (203 passed, 1 skipped)
- `npm run lint`
- type check and targeted fuzz regression

Fixes #1043


motdotla:
fixing here and will go out with next patch: https://github.com/motdotla/dotenv/pull/1056

PR #1049, opened 2026-09-12
fix: populate() crashes with TypeError when passed null

webdevelopersrinu:
`populate(process.env, null)` crashes with a raw `TypeError`. It should
throw the `OBJECT_REQUIRED` error, but `typeof null === 'object'` passes
the check. `populate(null, parsed)` crashes too — `processEnv` was never
checked, even though the error message says to check it.

I fixed the guard to catch null for both arguments. Added two tests.
Tests pass on my machine.


lbesecker195 (approved):
Verified: `typeof null === 'object'` slips past the existing check, and `processEnv` was never validated even though the error message names it. The new guard covers both, and the tests exercise both paths. LGTM.

PR #1050, opened 2026-09-13
test: make the test suite pass on Windows and run CI on windows-latest

kaanisthatyou:
On Windows, `npm test` currently fails 3 tests because they hardcode POSIX output:

- `deals with file:// path` and `deals with file:// path and debug true` expect `open 'file:///tests/.env'`, but on Windows Node reports the resolved path (`open 'C:\...\file:\tests\.env'`). The expected message is now taken from `fs.readFileSync` on the same path.
- `displays the injected env message without tips` expects `from tests/.env`, but `path.relative` returns `tests\.env` on Windows. The expected path is now built with `path.join`.

CI now also runs on `windows-latest`, so this stays green.

Tested: Windows 11 / Node 24.14.0 locally went from 187 passing and 3 failing to 190 passing (1 skipped, unchanged). The updated workflow passes on a fork for ubuntu-latest and windows-latest with Node 20, 22, 24 and 26 ([run](https://github.com/kaanisthatyou/dotenv/actions/runs/34762393677)).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01XvSbVkca8tLnsnVVNKUQHQ


motdotla:
nice Pr. thanks @kaanisthatyou !

PR #1054, opened 2026-09-18
fix: handle file URLs in config logging

jakezwang:
Passing a file `URL` as `path` loads the file but returns `ERR_INVALID_ARG_TYPE` when the injection message calls `path.relative()`. For a missing file, the logging error also replaces the original `ENOENT`.

Convert file URLs to paths when formatting that message. Tests cover a successful URL load, escaped spaces/percent signs in debug mode, and preservation of file-read errors.

Validation: `npm test` passes on Node 20, 22, 24, and 26 (346 assertions passed, 2 platform-specific skips each), including build, lint, and type checks.


motdotla:
nice fix. thanks for catching.

PR #1058, opened 2026-09-22
fix: keep assignments after a bare export line in the fast parser

SulimanAbdulrazzaq:
In the `{ fast: true }` parser, a line that holds only `export` makes the next `export KEY=value` line disappear. The classic parser keeps it.

```js
const src = 'export\nexport KEY=value\n'

dotenv.parse(src)                  // { KEY: 'value' }
dotenv.parse(src, { fast: true })  // {}
```

Trailing spaces or blank lines after the bare `export` behave the same way: `'export \n\nexport KEY=value\nNEXT=ok'` loses `KEY` and keeps `NEXT`.

### Cause

After `export`, the scanner skips whitespace to find the key, and that skip crosses line breaks. After a bare `export` line it lands on the next line's `export`, reads it as the key, finds no `=` or `:` after it, and the invalid-line branch then skips to the end of that line, taking `KEY=value` with it. The classic regex backtracks and matches the second line on its own.

### Fix

Remember where the `export` keyword ends, and when the key it led to has no separator, resume from the `export` line instead of the key's line. This is the same thing the invalid-line branch already does for the separator lookahead. Nothing changes when `export` is followed by a valid key.

### Tests

Two cases added to `fast parse matches classic for colon newlines and export keys` in `tests/test-parse-fast.js`. Both fail for the fast parser on `master` (5203712) and pass with the fix. `npm test` (build, lint, dts-check, tap) passes locally on Windows with Node 24: 1086 passed, 2 skipped (the POSIX signal test and the perf file).


motdotla:
thanks. great catch!

